Issue #9, opened 2015-08-13
FPS Limit


EtienneBruines:
When using the example given at the README, I'm getting a FPS of about 30-40k. 

It would be nice if we could limit this, since this wastes quite some CPU resources. The programmer could pass/set options somewhere, to set the limit to 30/60/90/120/unlimited whenever he wants. Preferably at runtime (because it'll reside in the game Settings). 

Looking at the code [here](https://github.com/paked/engi/blob/master/engi_glfw.go#L125-L132), it's obvious why the FPS is getting this high. 

What about something like this?

```go

	Wo.New()
	ticker := time.NewTicker(time.Duration(int(time.Second) / fpsLimit))

Outer:
	for {
		select {
		case <-ticker.C:
			responder.Update(Time.Delta())
			window.SwapBuffers()
			glfw.PollEvents()
			keysUpdate()
			Time.Tick()
			if window.ShouldClose() {
			    break Outer
			}
		}
	}
    ticker.Stop()
```

Not sure what the best practice would be if someone were to change the fpsLimit - we'd have to re-create the ticker.

paked:
Well looking at the exposed functions, we would have to recreate it. I personally am happy with this solution, it's not like people would recreate the ticker every frame.

EtienneBruines:
Would you prefer this to be handled as a static function, such as the `SetBackground` function? i.e. 
```go
func SetFPSLimit (fps int) {
    ....
}
```

Or tied to any running instance of some sort?

paked:
Static seems to be the only possible way to do it right now, without a massive restructure. :)

pedronasser:
LGTM

PR #10, opened 2015-08-14
Added support for FPS Limit. Currently the default is set to 120.

EtienneBruines:
As discussed in #9. Let me know if you have any remarks.

### DIFF
--- a/engi.go
+++ b/engi.go
@@ -4,7 +4,11 @@
 
 package engi
 
-import "github.com/ajhager/webgl"
+import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


EtienneBruines:
I may have made a mistake in making `case <-resetLoopTicker:` the second case. Don't know what this does when the first one keeps firing faster than it can process. Could be that the resetLoopTicker never gets its turn...

paked:
That looks OK to me. Only thing I would say is to implement it in the examples (or create an example to showcase its feature)

pedronasser:
LGTM.

Issue #14, opened 2015-09-30
Fix examples


paked:
It seems like the examples are broken from when better font support was added.

* [x] Pong
* [x] botmark.go
* [x] game.go
* [x] hello.go
* [ ] tilemap.go

EtienneBruines:
This commit https://github.com/paked/engi/commit/ff424f9414d2355bde697560f86831863f0f2e47 seems to be (partially) to blame. Referencing because it shows old behavior with the Fonts.

paked:
@EtienneBruines yes, I believe the problem is that we never got around to updating the examples when we made an API change (d'oh!). It is not necessarily a bug as such, as a working updated example can be found [here](https://github.com/Newbrict/OakTale).

EtienneBruines:
Looks like the font now sees https://github.com/paked/engi/blob/master/demos/pong/assets/font.png as a bad ttf file.

EtienneBruines:
Note that the example `falling.go` has a `DeathSystem` that listens for `ScoreMessage`s, but they are not defined, nor sent.

paked:
@EtienneBruines Was this fixed in your latest PR?

EtienneBruines:
As it wasn't a compiler-error, no, it wasn't. I wasn't sure at which point it should send `ScoreMessage`s.

paked:
@EtienneBruines makes sense. I'll add this to my TODO list.

paked:
Regarding this. I believe we can drop the `botmark` demo. It was a benchmark, and those are being worked on in #40

EtienneBruines:
:thumbsup:

EtienneBruines:
This is being continued on #161 and #162. 

Closing this issue as it does not contribute anything further.

PR #15, opened 2015-10-01
Updated pong demo and font

EtienneBruines:
Pong demo now works with the Font API changes. Also updated
the ScoreSystem, to only update the score Texture when
the score actually changed since last time. Added mutex
to ensure thread-safety. 

Updated font.go to return an error value for Create, so one
can handle a font that couldn't be loaded (i.e.: load another/crash).

Should fix the first item within #14.


paked:
Wow! This is awesome. I'm having a busy night, I will get around to helping out sometime over the next few hours :)

EtienneBruines:
Don't feel bad for having a busy night. It's open source - everyone helps whenever he/she wants to :-)

PR #17, opened 2015-10-22
Updated the Camera API

EtienneBruines:
## Camera API
Objects are now positioned/drawn where they should be. The
`glfw` part worries about where the camera is, and where the
objects should be rendered.

Added some tests to ensure its behavior doesn't change over time.

Removed the `Clamp` function in the math file, because it's
probably better to use the `mgl32` package for things like that. No
need to define it twice.

Renamed the `gl` variable in engi, because having both `gl` and `Gl`,
referencing to the exact same object, doesn't make any sense.

## KeyboardScroller
I also added a `KeyboardScroller`. It's a `Systemer` people can add using 
```go
game.AddSystem(engi.NewKeyboardScroller(scrollSpeed, engi.W, engi.D, engi.S, engi.A))
```
And then everything works out-of-the-box. No need to create a separate `Entity` for it. 

## EdgeScroller
Much like the `KeyboardScroller`, we now have a `EdgeScroller` which everyone can use:
```go
game.AddSystem(engi.NewEdgeScroller(scrollSpeed, edgeMargin))
```
Then all you need to do, is move your cursor to the edge of the window, to scroll :-).

### DIFF
--- a/assets.go
+++ b/assets.go
@@ -252,7 +252,6 @@ attribute vec2 in_TexCoords;
 
 uniform vec2 uf_Projection;
 uniform vec3 center;
-//const vec3 center = vec3(-1.0, 1.0, 1.0);
 
 varying vec4 var_Color;
 varying vec2 var_TexCoords;
@@ -260,9 +259,9 @@ varying vec2 var_TexCoords;
 void main() {
   var_Color = in_Color;
   var_TexCoords = in_TexCoords;
-  gl_Position = vec4(in_Position.x /  uf_Projection.x - center.x,
-				 	 in_Position.y / -uf_Projection.y + center.y,
-										 0.0, center.z);
+  gl_Position = vec4(in_Position.x /  uf_Projection.x - center.x - 0.5,
+				 	 in_Position.y / -uf_Projection.y + center.y + 0.5,
+										 0, center.z);
 }`
 
 var batchFrag = `
--- a/assets.go
+++ b/assets.go
@@ -106,20 +106,20 @@ type Image interface {
 }
 
 func LoadShader(vertSrc, fragSrc string) *webgl.Program {
-	vertShader := gl.CreateShader(gl.VERTEX_SHADER)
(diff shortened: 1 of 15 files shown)


paked:
Is there any chance we could get an example of each `*Scroller` in the `demos` folder? Apart from that, LGTM.

EtienneBruines:
One demo for each, or one demo that includes both?

paked:
@EtienneBruines one demo for each

EtienneBruines:
Will have it ready some time tomorrow evening (so that'll be Saturday for you ;-) ).

paked:
@EtienneBruines Awesome. Thanks for the contribution, its really awesome!

matiwinnetou:
awesome stuff :+1:

EtienneBruines:
@paked I had some time to kill, so here are two demos. 

As you'll see in the demos, the `Camera` behavior isn't perfect just yet (it doesn't let you zoom down all the way).

paked:
@EtienneBruines pinged you on slack... Having troubles getting this to compile:

```
# github.com/paked/engi
../../batch.go:117: Gl.Uniform3f undefined (type *webgl.Context has no field or method Uniform3f)
```

EtienneBruines:
My bad. https://github.com/paked/webgl/pull/1

paked:
@EtienneBruines I'd really like to get the padding issues fixed up before merging this in. Do you have the bandwidth for this, or should I merge in and investigate myself?

paked:
Actually it looks like this may not be a problem in the `Camera` API

EtienneBruines:
Most-likely not in `camera.go`, but the code that's listed in `assets.go` and `batch.go` is certainly involved. 

(I see my response went missing; I said that I've got no idea how to fix the problem; has something to do with that Shader code (C) )

paked:
@EtienneBruines I'll take a look later today.

Thanks for the contribution!
taking_content

PR #20, opened 2015-10-23
Fix examples

paked:
@EtienneBruines I'd appreciate it if you could review these changes :)


paked:
This should hopefully make it more clear how engi is designed to be used

paked:
Thanks for the contribution me!
what

Issue #27, opened 2015-10-25
Feature Request: Menu


EtienneBruines:
It'd be nice if someone pressed something like `ESC` (or something else), it'd pop-up some kind of menu. We all know it; it asks us if we want to continue, change settings, exit, save, load, etc.

But it also stops everything else in the game -- depending on which game you're playing. 
## Proposition:

``` go
width := 600
height := 800
pauseWorld := true

menuEntity := engi.CreateMenu(width, height, pauseWorld)
```

And this `Entity`, would then contain a `SpaceComponent` of given size, and one could add some additional components (i.e. `RenderComponent`) to do the markup themselves. The `pauseWorld` part is why I'd love to see this implemented (by me or someone else). 
## Additional proposition (which you don't have to like):

``` go
pauseWorld := true
items := map[string]func() {
    "Continue" : func() { /* do nothing */},
    "Save" : func() { /* save the game */},
    "Exit" : func() { /* exit the game */},
}

menuEntity := engi.CreateMenu(pauseWorld, items)
```

Then `engi` would just create a menu with these options, possibly with some styling options for the user. 

Not sure how this fits in the whole ECS-paradigm, but it'd be a nice feature to have. And this issue would be a great place to discuss (or summarize discussions).

paked:
Awesome. We definitely need some sort of pause/menu feature. My only concern would be making the menu too specific and not flexible enough for general use. IMO we should view the pause menu as a group of `Entity`s which represent a UI + pause logic.

Regarding your second proposal, I don't quite understand why you wouldnt just use an interface?

```Go
type XYZ interface {
   Continue()
   Save()
   Exit()
}
```

Also, perhaps implementing `Save` logic into the engine is going a bit too far. I'm happy for a discussion on this though.

EtienneBruines:
So it should have some kind of `PauseSystem`, which some `Entity` would `require`?

w.r.t. the second proposal:
The idea was the `map`, not those specific entries of the map. It would create a menu with the strings as buttons (or w/e), and execute the function on select.

paked:
Yes. A `PauseSystem` would be good. We could make use of the `Message`-ing features, to summon the `PauseSystem`.

@EtienneBruines I think it would be best to allow each `Entity` to handle its own behaviours.

EtienneBruines:
A `PauseSystem` with the `Message`-ing features would work. 

[...]

PR #33, opened 2015-10-28
Implemented PauseSystem and updated Message

EtienneBruines:
**[BREAKING CHANGE]**
`MessageManager.Dispatch` now only requires the `Message`, no `Name` `string`.
This is because this information is already in `Message.Type()`

`Handler` now is no longer `func(i interface{})`, but `func(msg Message)`,
since we know it'll be a message - it's best to use type safety.
It's also renamed to `MessageHandler`, to be more specific.

Fixed the examples so it didn't break anything.

Added demo for the `PauseSystem`. Should go a long way with #27.


paked:
LGTM once we have established this is the messaging model we want to use

cc: @EtienneBruines @matiwinnetou

EtienneBruines:
w.r.t. the `PauseSystem`, or in general?

paked:
@EtienneBruines In general. I personally think this is the best model...

EtienneBruines:
Removed the log/debug message. Some TODOs can't change until #35 is fixed.

EtienneBruines:
Gotta say. Trying to refactor `Responder` has made me think about the design.

There's a separate `PauseSystem`, to which a message will be sent. This system then, alters the `World`. 

The only things that can send such `Message` to the `PauseSystem`, are other `System`s. Why don't we let them directly manipulate the `World`? It'll still be a `System` that modifies the `World`.

paked:
@EtienneBruines This seems interesting, although I'm not quite sure I understand what you are saying

EtienneBruines:
Why are we requiring the user-made `System`, to send a `Message` to our `PauseSystem`, such that our `PauseSystem` can alter the `World` (pause it), 

when that user-made `System` could alter the `World` (pause it) itself, directly?

paked:
@EtienneBruines that makes sense. So you're saying we make pausing behaviour a first class citizen?

EtienneBruines:
Wouldn't hurt. The only difference between that, and the current implementation:
* Now `World` has a unexported variable `isPaused`;
* Now `PauseSystem` can alter that value directly, when it receives a message;
* Then, a simple `World.Pause()` function would suffice - no need for messages. 

Since it's directly *changing the world*, if it's a system, it shouldn't matter if we programmed the system, or if they did?

paked:
I think perhaps it is better design for us to "eat our own dog food" and build internal Systems following the pattern we expect others too.

EtienneBruines:
So, remove `PauseSystem`, and create some kind of `Pause` function on `World`?

paked:
No. Build a `PauseSystem` for others to use, which makes use of a private function `pause()` on `World`.

EtienneBruines:
Roger that. Then I believe this PR does exactly that. (perhaps not a private `pause()` function, but alters the `isPaused` variable).

matiwinnetou:
I would agree with @paked but maybe over time I may change my mind. Now I think we should really use ECS where we can but maybe over time we will find which things work, which don't.

paked:
OK LGTM then.

PR #34, opened 2015-10-28
Renderables not regions...

matiwinnetou:
This should support animations of vector graphics.


EtienneBruines:
LGTM.

paked:
Thanks for the contribution!

PR #39, opened 2015-10-30
Added AudioSystem / Changed to CameraSystem

EtienneBruines:
AudioSystem:
* First attempt with audio and OpenAL.
* Also works 3d-ish, including distance.
* We have to convert Stereo16 to Mono16, because only
mono sounds are affected by ditance.

CameraSystem:
* Updated Camera to CameraSystem

General:
* Changed all demos, so they won't be affected

Note: this is initial support for Audio; lots of improvements to be made, but it's nice to have something.


paked:
Awesome contribution @EtienneBruines!

EtienneBruines:
Still have to force-push some changes, so don't push just yet, but I'm waiting for that so I won't break comments (not sure if it would, but would be a shame if it did).

EtienneBruines:
Everything is now according to the changes discussed. 

One thing I didn't address, is the `Playing` `State` w.r.t. the GoF. This is mostly because I'm not sure what I should be changing. 

It looks good to me, but as this is my PR, someone else should say/approve it :-).

paked:
So, I'd be happy to have GoF in another PR.

matiwinnetou:
GoF State pattern could be an overkill. One would have to play and check if this is not over engineered for a simple use case.

paked:
@matiwinnetou does this LGTY?

matiwinnetou:
yes, LGTM

paked:
Thanks again @EtienneBruines!

Issue #40, opened 2015-10-30
Benchmarking / Non-GLFW solution


EtienneBruines:
## What

Would be nice to have a way to "disable" GLFW renderings, and just test the "logic" behind it all. In other words: I want to be able to benchmark `engi`, using `go test`. 

## Why
This makes automated tests easier, as well as benchmarking a certain solution (i.e. `Message`-usage) - in order to improve its performance. 

------

## Tasks (and possibly more to add):

- [ ] allow to remove FPS limit (only then, can we actually benchmark)
- [x] fix termination/crash of GLFW when within `go test` operations

paked:
I agree. I really would like to have a headless mode too. It would make
writing multiplayer games easier too!

EtienneBruines:
I'm not that experienced in multiplayer games; how would doing this make it easier?

(the only thing I can think of, is having the actual game "played" at the server (which would require the headless mode), and have the clients just have a representation of what's going on?

EtienneBruines:
Shall I split this issue into two?

* Allow GLFW in tests
* Create a headless mode

paked:
@EtienneBruines Not really much point now...

EtienneBruines:
I guess I just fixed both of these, in one commit ... Thought it'd be more work. Apologies!

paked:
@EtienneBruines Don't apologise! Great job!

@EtienneBruines Yes. Regarding that is how multiplayer games (usually) work. Server runs the "master copy" of the game, clients run a "local copy" and forward their keystrokes etc to server.

PR #41, opened 2015-10-31
Created headless mode

EtienneBruines:
Should be halfway for #40.

One can now choose to run (hopefully the same) code in
headless mode; no `GLFW` calls will be made.

`System`s can now also be set to `SkipOnHeadless`, so these `System`s
will get skipped all together.

Added headless-pong demo to see it in action.


matiwinnetou:
This is for testing I guess?

EtienneBruines:
Mostly testing/benchmarking, yes

paked:
SO COOL!

paked:
Thanks for the contribution!

PR #47, opened 2015-11-01
we generate id when creating entity not later

matiwinnetou:


utils.go
```diff
@@ -0,0 +1,17 @@
+package engi
+
+import (
+	"crypto/rand"
```
EtienneBruines: Any reason why we're not using `math/rand`? (should be faster)

paked: Yeah. `crypto/rand` is more *random*... But security isn't really what we need in this situation (speed is). `math/rand` should suffice.

paked:
No problem @matiwinnetou!

PR #48, opened 2015-11-01
Added ability to remove entities from the world

matiwinnetou:
Exist = false may not be performant way to remove enitites.

Imagine a game when there are bullets, lots of small short lived objects. If they are all stored in the world the game would run very slowly after a while not to mention memory footprint.

### DIFF
--- a/demos/animation/anim.go
+++ b/demos/animation/anim.go
@@ -20,6 +20,7 @@ type GameWorld struct {
 }
 
 func (game *GameWorld) Preload() {
+	game.New()
 	engi.Files.Add("assets/hero.png")
 	game.STOP_ACTION = &engi.AnimationAction{Name: "stop", Frames: []int{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}}
 	game.RUN_ACTION = &engi.AnimationAction{Name: "run", Frames: []int{16, 17, 18, 19, 20, 21}}
--- a/demos/audio/audio.go
+++ b/demos/audio/audio.go
@@ -51,6 +51,7 @@ type GameWorld struct {
 }
 
 func (game *GameWorld) Preload() {
+	game.New()
 	engi.Files.Add("assets/hero.png")
 	engi.Files.Add("assets/326488.wav")
 	engi.Files.Add("assets/326064.wav")
--- a/demos/falling/falling.go
+++ b/demos/falling/falling.go
@@ -14,7 +14,8 @@ type Game struct {
 	engi.World
 }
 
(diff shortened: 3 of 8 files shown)


EtienneBruines:
Good to be able to remove `Entity`s. 

Some thoughts:
* would it be faster to use a `map`, instead of a list? (we could benchmark this? ) - not really related to the code you just wrote;
* would it be faster if we did some "garbage collection" of our own? (i.e. delete all non-exiting elements every `x`th frame? or every `y`th garbage item?)

Note that removing one element (each time you need it) might be costlier than not doing it on-the-fly. 

For now, it's nice that we can actually remove entities :-).

paked:
I agree. Anything else seems like a premature optimisation. Lets wait until we have #40 happening with tests before we do that.

paked:
I'm wondering. Would it make sense to write a test for this? I figure we will do it eventually, why not start now?

Thoughts @EtienneBruines @matiwinnetou?

EtienneBruines:
Not sure what we would be testing. Benchmark would make more sense?

matiwinnetou:
@paked @EtienneBruines this is core code so test would make sense (to catch a regression). As far as benchmarking, I dunno, perhaps it is easier to operate on a vector then to recreate arrays like crazy. Map I am not sure why, do we need a key value map?

matiwinnetou:
@EtienneBruines stupid me, map will be perfect for this use case, but only if we merge uuid generator for entities first. Keys will be entity ids

paked:
Sounds good to me

matiwinnetou:
@paked @EtienneBruines review new code please?

EtienneBruines:
Code LGTM. Haven't had the time to test it yet, but I assume it works ;-).

paked:
:+1: for adding in a mutex.

matiwinnetou:
I would disagree, it only makes sense if we make a decision to multithread. A lot of UI engines are single thread only and this is a general question if we want to answer ourself if we want the engine to use the power of multiple cpus at the higher complexity to reason about the code and deal with whole topic of potential nightmare of concurrency bugs.

EtienneBruines:
I do get your point as to "worry about it when we start to multithread", since we have to look at a lot of code to make that happen.

paked:
@matiwinnetou good point. Thinking about it we actually don't use any concurrent features at the moment. Until we do, I do not see the point of adding this in.

EtienneBruines:
LGTM.

paked:
Thanks for the contribution

PR #51, opened 2015-11-04
Gamesetup

EtienneBruines:
Not yet ready to merge just yet, because it broke the `Scroll` part, but the rest is OK.


EtienneBruines:
This is not going as planned ; it somehow has an old commit, which should not be merged ...

PR #50, opened 2015-11-04
Benchmark

EtienneBruines:
Added some benchmarks (not all of which are as useful), to benchmark certain `System`s.

Please review @matiwinnetou @paked, as I did some refactoring of the `Open` and `run` functions, which possibly could be done in a better manner. 

Some output:
```
➜  engi git:(benchmark) go test -bench .                                     
PASS
BenchmarkEmpty-8              	10000000	       116 ns/op
BenchmarkSystem10-8           	 5000000	       309 ns/op
BenchmarkSystem1000-8         	  100000	     17569 ns/op
BenchmarkEntity10-8           	10000000	       179 ns/op
BenchmarkEntity1000-8         	  300000	      3961 ns/op
BenchmarkCollisionSystem10-8  	   50000	     39614 ns/op
BenchmarkCollisionSystem100-8 	     500	   4015496 ns/op
BenchmarkCollisionSystem1000-8	       3	 446231737 ns/op
ok  	github.com/paked/engi	15.830s
```

This also allows for some profiling:

```
      flat  flat%   sum%        cum   cum%
    1560ms 16.27% 16.27%     2050ms 21.38%  runtime.mallocgc
    1080ms 11.26% 27.53%     2190ms 22.84%  runtime.mapaccess2
     820ms  8.55% 36.08%     8110ms 84.57%  github.com/paked/engi.(*CollisionSystem).Update
     670ms  6.99% 43.07%     4730ms 49.32%  github.com/paked/engi.(*Entity).GetComponent
     530ms  5.53% 48.59%      530ms  5.53%  reflect.ValueOf
     400ms  4.17% 52.76%      400ms  4.17%  runtime.aeshash64
     400ms  4.17% 56.93%      800ms  8.34%  runtime.interhash
     350ms  3.65% 60.58%      350ms  3.65%  runtime.mSpan_Sweep.func1
     340ms  3.55% 64.13%      340ms  3.55%  reflect.Value.Elem
     210ms  2.19% 66.32%     2260ms 23.57%  runtime.newobject
     190ms  1.98% 68.30%      670ms  6.99%  reflect.Value.Set
     190ms  1.98% 70.28%      280ms  2.92%  reflect.Value.assignTo
     180ms  1.88% 72.16%      180ms  1.88%  runtime.ifaceeq
     160ms  1.67% 73.83%      160ms  1.67%  reflect.Value.Type
     150ms  1.56% 75.39%      500ms  5.21%  runtime.heapBitsSweepSpan
[...]

PR #54, opened 2015-11-04
Tuned Performance

EtienneBruines:
### CollisionSystem
Performance increased about `10-15%`.

### Component
Performance increased about `20-25%`. 

(I know the branch has an incorrect name ;-) )

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -6,7 +6,7 @@ import (
 
 type Entity struct {
 	id         string
-	components map[reflect.Type]Component
+	components map[string]Component
 	requires   map[string]bool
 	Exists     bool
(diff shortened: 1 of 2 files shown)


paked:
This LGTM.

paked:
This doesn't depend on any changes from #50 does it?

EtienneBruines:
No, it does not. 

I did use #50 to come up with the performance measurements, but these changes are separate.

EtienneBruines:
Merging because of the `LGTM` - and this way I can also change the other PRs to reflect changes.

paked:
Good call.

PR #58, opened 2015-11-06
Fonts can now be preloaded

EtienneBruines:
Updated demo to use new feature.

Nothing special, but very nice to use.


paked:
Thanks for the contribution!

PR #61, opened 2015-11-07
Fixed bug in RenderSystem

EtienneBruines:
Whenever an item requiring the RenderSystem, didn't have a SpaceComponent, engi would crash.
This is fixed now.


EtienneBruines:
No need to debate long about this. Obvious mistake, obvious fix.

paked:
:+1:

PR #68, opened 2015-11-07
Fix audio example being too complex

paked:
The previous audio example contained animation and movement.

IMO all examples should be their own little feature, and not combine the two together (Unless they are a sample game, like pong).

### DIFF
--- a/demos/audio/audio.go
+++ b/demos/audio/audio.go
@@ -4,95 +4,26 @@ import (
 	"github.com/paked/engi"
 )
 
-type moveSystem struct {
-	*engi.System
-}
-
-func (moveSystem) Type() string {
-	return "moveSystem"
-}
-
-func (ms *moveSystem) New() {
-	ms.System = engi.NewSystem()
-}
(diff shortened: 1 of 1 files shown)


paked:
Maybe it would be beneficial to have both examples?

EtienneBruines:
Perhaps it should also show that audio can be "background audio", and "not background audio"? 

The reason it allowed for scrolling, is because the more you zoom out, the less you hear the weird man, but the bird stay the same.

paked:
@EtienneBruines Ah. I see your point. I'll add those back in, but not the animation or movement.

EtienneBruines:
:+1:

matiwinnetou:
+1 on complexity. Such combo things we could demo in an example game but those small demos should help in writing tutorials.

paked:
@EtienneBruines I think we should keep this example as it is. But have a second example with what you previously mentioned.

I'll do this in a separate PR.

Merge when ready.

PR #70, opened 2015-11-07
Fixed HUDGround ratio

### LINKED ISSUE #66: HUDGround ratio
The current `HUDGround` `PriorityLevel` does not have a correct perception of pixels. 

If you draw something with width `800`, and have a `1600px` monitor, you will see that the `Entity` covers the entire screen, not half. 

Even though it's not hard to just do /2 in the users' code, this should be fixed.

EtienneBruines:
Fixes #66. Also improved performance because I simplified the formula.


paked:
LGTM.

PR #73, opened 2015-11-08
Added ComponentFast + Benchmarks

EtienneBruines:
It shows that using `ComponentFast` a lot faster.

```
BenchmarkComponent-8          	   20000	    155262 ns/op	   16200 B/op	    1002 allocs/op
BenchmarkComponentDouble-8    	   10000	    266979 ns/op	   24200 B/op	    2002 allocs/op
BenchmarkComponentFast-8      	   50000	     70341 ns/op	    8200 B/op	       2 allocs/op
BenchmarkComponentFastDouble-8	   30000	     86215 ns/op	    8200 B/op	       2 allocs/op
```

* The benchmark is: how long does it take for one iteration / frame to run, with 1000 `Entity`s tied to one `System`, which either uses `GetComponent`, or `ComponentFast` to get its `Component`s. 
    * note that this means that the `ns/op` also includes other stuff, unrelated to `GetComponent`, such as iterating over all `Entity`s. 
* The `Double` postfix means that it tries to parse two components, i.e.: `SpaceComponent` and `RenderComponent`.

~~The syntax of the `ComponentFast` is not yet what I'd want, but I'm not yet sure what I want it to be.~~

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -75,9 +75,13 @@ func (a *AnimationSystem) Update(e *Entity, dt float32) {
 	var (
 		ac *AnimationComponent
 		r  *RenderComponent
+		ok bool
 	)
 
-	if !e.Component(&ac) || !e.Component(&r) {
+	if ac, ok = e.ComponentFast(ac).(*AnimationComponent); !ok {
+		return
+	}
+	if r, ok = e.ComponentFast(r).(*RenderComponent); !ok {
 		return
 	}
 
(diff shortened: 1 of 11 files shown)


EtienneBruines:
@paked shall I also make a commit which updates all usage of `GetComponent` in the `engi` core, to `ComponentFast`?

paked:
@EtienneBruines SGTM

EtienneBruines:
@paked I have changed the syntax. 

```go
var (
    space *SpaceComponent
    ok bool
)

if space, ok = entity.ComponentFast(space).(*SpaceComponent); !ok {
    return
}
```

This way, we don't have to hardcode the `Type()` string twice. 

Please review @paked / @matiwinnetou if you want; can me merged when you're ready.

paked:
Yes. This. :+1: (will review in a few)

PR #76, opened 2015-11-09
Added support for RGBA.

EtienneBruines:
One should also update the webgl package.


paked:
Thanks for the contributions @EtienneBruines!

PR #83, opened 2015-11-10
Fixed empty entries in World.Entities

EtienneBruines:
`append` already increases length and capacity, so if you create something with length `100`, and then `append` 100 things to it, then it ends up with a length of `200`.

### DIFF
--- a/world.go
+++ b/world.go
@@ -56,7 +56,7 @@ func (w *World) AddSystem(system Systemer) {
 }
 
 func (w *World) Entities() []*Entity {
-	entities := make([]*Entity, len(w.entities))
+	entities := make([]*Entity, 0, len(w.entities))
 	for _, v := range w.entities {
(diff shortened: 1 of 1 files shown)


matiwinnetou:
Intuitively I would agree with @EtienneBruines  but at least in Java world JVM is doing lots of advanced optimisations (JIT) while running.

paked:
I'm very hesitant, as this seems like an over optimisation... Have we even got any games that are pushing struggling without this yet?

EtienneBruines:
Well, *something* needs to fix, because the current implementation has empty entries. Which one it becomes, is less of an issue. Both variants work. 

But as this is something that's done 60-120 times per second, performance is something we should at least think of.

paked:
Good point. This LGTM...

paked:
Thanks for the contribution!

PR #86, opened 2015-11-13
Hotfix draw performance

### LINKED ISSUE #85: Fix performance issues
I don't know *exactly* why (yet), but [this code](https://github.com/EtienneBruines/bcigame) runs (after selecting "New Game", and adding 100 entities on-screen), at a mere 17 FPS. 

I'm not saying I want to reach 1,8 million fps again (which we'll never reach with rendering), but looking at the things I'm doing, I feel like this should be fixed. 

The go profiler says it's mostly `external code` :disappointed: . 

------

* The problem definitely has something to do with `engi.RenderSystem`. 
* More precisely, this function call (line `119` in `batch.go`):

```go
Gl.BufferSubData(Gl.ARRAY_BUFFER, 0, b.vertices)
```

* ~~`b.vertices` is of length `200.000` (maximum length) every time. Not sure why that is the case, but it sure is related to the problem.~~
* I am rendering 100 *different* entities; that might be related (a lot of calls to that function)
* That function is called about 30 times per frame; which could be worse, but as it requires `0.1ms` to run, that's too much to handle.

EtienneBruines:
Fixes #85.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -116,7 +116,7 @@ func (b *Batch) flush() {
 	Gl.Uniform2f(b.ufProjection, b.projX, b.projY)
 	Gl.Uniform3f(b.center, cam.x/Width(), cam.y/Height(), cam.z)
 
-	Gl.BufferSubData(Gl.ARRAY_BUFFER, 0, b.vertices)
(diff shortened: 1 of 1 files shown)


paked:
Thanks for the contribution @EtienneBruines!

PR #89, opened 2015-11-25
Added RunInParallel

EtienneBruines:
`Systemer` now includes a function `RunInParallel`, which indicates whether or not the `Entity`s should be processed in Parallel. 

`CollisionSystem` only benefits from `RunInParallel`, if it has more than 40 entities. This is because `CollisionSystem` is quite fast when there's less than 40 entities, and the overhead of starting goroutines is more. 

Each `System` has its own turning point, which can be defined without much effort (see code).

### DIFF
--- a/collision.go
+++ b/collision.go
@@ -64,6 +64,10 @@ func (cs *CollisionSystem) New() {
 	cs.System = NewSystem()
 }
 
+func (cs *CollisionSystem) RunInParallel() bool {
+	return len(cs.entities) > 40 // turning point for CollisionSystem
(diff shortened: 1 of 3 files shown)


EtienneBruines:
Merged because the current state of `master` was just unacceptable (unstable). When rendering 20 objects, 2-4 of them were gone due to concurrency issues. 

Note that everything is `serial` (not-concurrent) by default now. This ensures that everything runs the way it should.

paked:
:+1:

PR #94, opened 2015-11-25
Made World available in System.New

### LINKED ISSUE #93: System New does not have access to system.World
The `system.World` variable is being set after the `system.New()` function is being called. 

As such, one does not have access to the `World` from the `system.New()` function (null-reference). This should be fixed by simply switching two lines: first calling `SetWorld`, then calling `New()`.

EtienneBruines:
Fixes #93.


EtienneBruines:
Does not work, awaiting fix.

PR #107, opened 2015-11-26
Sorting Systems in World

### LINKED ISSUE #105: Sorting System by Priority
Each `Systemer` has a `Priority()` function, which returns an `int`. 

We should sort the list of `Systemer`s in `World`, after adding one, by that `Priority` value. It was probably added with that idea, but it was never implemented.

EtienneBruines:
When adding a `System` to `World`, it's appended
to the end, and then a quicksort algorithm is used
to sort it according to `System.Priority()`.

Fixes #105.


paked:
LGTM

PR #118, opened 2015-12-08
Added Scenes

### LINKED ISSUE #77: Scenes in ECS
With the term `Scene`, I mean something that fills the screen. Such as "level 1 in pacman", or "main menu", or "settings", or "high scores". 

How should one handle this / switch between them, in ECS / engi? 

My original thought was to add / remove `System`s on-the-fly, indicating which would be active and which would not. I am not sure if this is the best/correct way to go, or if there's another (more idiomatic) approach?

EtienneBruines:
**BREAKING CHANGE**

Also added Scenes Demo and updated all other demos. Fixes #77. 

This is ready to be merged and reviewed by @paked (and others).


EtienneBruines:
Updated.

PR #129, opened 2015-12-14
Fixed some global variables wrt Scene

### LINKED ISSUE #126: Scenes and Messages
Having multiple instances of the same type of `System`, when the `System.New(*ecs.World)` function registers it at our `engi.Mailbox`, messages get dispatched twice.

We should have one `Mailbox` per `Scene`.

### LINKED ISSUE #128: CameraSystem creation within RenderSystem creation
Consider these lines of code (currently in `master`), within the `New` function of `RenderSystem`:

```go
	// Set cameraSystem if doesn't already exist
	if cam == nil {
		cam = &cameraSystem{}
		w.AddSystem(cam)
	}
```

Looks OK, on first sight. However, with the introduction of `Scene`s, this isn't correct anymore (as we're using a global `cam` variable.

EtienneBruines:
Fixes #126 and fixes #128.

Global variables are now updated on Scene creation (whenever
we initialize a new World instance). These values are saved
in a sceneWrapper, so we don't always have to reinitialize.

CameraSystem is now set on Scene creation, since this was
hard to do within RenderSystem (as it is Scene-specific).

Ready for review @paked.

### DIFF
--- a/ecs/world.go
+++ b/ecs/world.go
@@ -84,6 +84,15 @@ func (w *World) Systems() []Systemer {
 	return w.systems
 }
 
+func (w *World) HasSystem(systemType string) bool {
+	for _, s := range w.systems {
+		if s.Type() == systemType {
+			return true
+		}
+	}
+	return false
+}
+
 // Update is called on each frame, with dt being the time difference in seconds since the last Update call
 func (w *World) Update(dt float32) {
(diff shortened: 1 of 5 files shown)


paked:
merging because breaking

paked:
good job @EtienneBruines

Issue #120, opened 2015-12-08
Update to luxengine/math


paked:
[luxengine/math](https://github.com/luxengine/math) has an implementation of the stdlib's `math` package, but using `float32`s... This would make our math code *a lot* nicer, removing the constant type casts.

EtienneBruines:
Never noticed that we did so much conversions. We should change it, yes.

Issue #21, opened 2015-10-24
Remove Legacy Code


paked:
Throughout the codebase there are various pieces of legacy code. These need to be removed.

* `Sprite`
* `Drawable` (or `Renderable`)

EtienneBruines:
Can I ask what's wrong with `Drawable`, and how it should look instead?

paked:
@EtienneBruines It fulfils the same purpose as `Renderable`, although... Perhaps we should use it... Renderable is quite ugly and relies on moving state.

EtienneBruines:
@paked Using either one of those would indeed be better; as they both sound like they're doing the same.

paked:
@EtienneBruines I personally think `Drawable` is the better design (not sure why I wrote it down then...)

cc: @matiwinnetou

matiwinnetou:
very useful info, thx. Helps me to learn the current status of the project.

paked:
@matiwinnetou Do you have any opinions on `Drawable` vs `Renderable`?

matiwinnetou:
@paked At the moment sadly no, as I understand still too little the difference. In general, however, from my experience I can say that one should avoid state like fire. It is not always possible though or complicates code a lot. You mentioned that Drawable has no state where as Renderable assumes state.

In our ECS app, which believe or not is actually web application. (We created own ecs-web framework in Java 8 we have even various Worlds, like ImmutableWorld, RequestWorld (mutable world), to make this separation clear. In our design worlds form a chain, where a world can have a parent attached.

paked:
@EtienneBruines What about you? (`Drawable` vs `Renderable` )

EtienneBruines:
I do think the `Drawable` is much clearer. Since there'll be a `Shader` for each `PriorityLevel` (which people can add themselves), all we need from whatever we want to draw, is the `Texture`, and size/location. 

I have no idea what the `View() (float32, float32, float32, float32)` function is supposed to do though.

PR #134, opened 2015-12-25
Changed Rendering

EtienneBruines:
Currently I've done quite a bit, and I think it is ready for review. 

### Shaders
* [x] HUD shader
* [x] User-defined shaders

### Rendering stuff
* [x] `Font` rendering
* [x] `Spritesheet` rendering - having *one* `Texture` in memory
* [x] `Region` rendering
* [x] `Level` rendering
* [x] ~~Consistently using `Renderable`~~ Replaced `Renderable` by `Drawable`. (See #21). 
* [x] Keep the `Scale` in mind, and allow it to be edited
* [x] Make sure `headless`-mode stays intact

### ECS
* [x] Basic functionalities of `ecs.Entity` (removed some of them)

### Engi startup
* [x] Refactor the way `Open`/`OpenHeadless` works, with a `RunOptions` struct
* [x] Allow `SwapInterval` to be user-defined
    * [x] + on runtime

### Cleaning up
* [x] Remove the `sketch` file, which was just to brainstorm
* [x] Ensure demos don't break
* [x] Replaced `float64` `math` by `float32` `github.com/luxengine/math` (see #120)


paked:
:+1: Will review when I have woken up a bit more :)

matiwinnetou:
wow, you rock Etienne

animation.go
```diff
@@ -1,12 +1,9 @@
-// Copyright 2014 Harrison Shoebridge. All rights reserved.
```
paked: Good call :)

render.go
```diff
@@ -49,32 +41,158 @@ func (renderChangeMessage) Type() string {
...
+		return
+	}
+
+	ren.bufferContent = ren.generateBufferContent()
+
+	ren.buffer = Gl.CreateBuffer()
+	Gl.BindBuffer(Gl.ARRAY_BUFFER, ren.buffer)
+	Gl.BufferData(Gl.ARRAY_BUFFER, ren.bufferContent, Gl.STATIC_DRAW)
+
+	// TODO: ask why this doesn't work
```
paked: // TODO

paked:
LGTM.

paked:
Keep in mind, I've completely forgotten everything about OpenGL... Perhaps you should get someone else to review that part?

shaders.go
```diff
@@ -0,0 +1,327 @@
...
+		s.flush()
+	}
+}
+
+func (s *HUDShader) flush() {
+	Gl.BindBuffer(Gl.ARRAY_BUFFER, s.coordBuffer)
+	Gl.BufferData(Gl.ARRAY_BUFFER, s.coordBufferContent, Gl.DYNAMIC_DRAW)
+	Gl.VertexAttribPointer(s.inCoordinates, 3, Gl.FLOAT, false, 12, 0)
+
+	Gl.DrawElements(Gl.TRIANGLES, 6*s.drawCount, Gl.UNSIGNED_SHORT, 0)
```
EtienneBruines: `s.drawCount` can be 0, so we should check before calling these functions.

demos/headless/headless.go
```diff
@@ -266,7 +267,7 @@ func (c *ScoreSystem) Update(entity *ecs.Entity, dt float32) {
 		c.upToDate = true
 		c.scoreLock.RUnlock()
 
-		render.Display = basicFont.Render(render.Label)
+		render.SetDrawable(basicFont.Render(render.Label))
```
matiwinnetou: if Render is a component, then why not assign drawable to a public instance variable?

demos/hide/hide.go
```diff
@@ -26,8 +24,8 @@ func (game *GameWorld) Setup(w *ecs.World) {
 	render := engi.NewRenderComponent(texture, engi.Point{8, 8}, "guy")
 	collision := &engi.CollisionComponent{Solid: true, Main: true}
 
-	width := texture.Width() * render.Scale.X
-	height := texture.Height() * render.Scale.Y
+	width := texture.Width() * render.Scale().X
```
matiwinnetou: why is it a method now and before was a public property access? To me it makes more sense that components in ECS have simply public access to variables.

demos/headless/headless.go
EtienneBruines: Because the `SetDrawable` function does two things:
* Set the variable value
* Update the OpenGL Buffers (i.e.: how big is this texture, what are the X,Y coords for each corner of the texture, etc.)

The latter is why there's a `SetDrawable` function, because forgetting to do that, will result in weird behavior.

demos/hide/hide.go
```diff
@@ -67,6 +65,10 @@ func (c *HideSystem) Update(e *ecs.Entity, dt float32) {
 }
 
 func main() {
-	World = &GameWorld{}
-	engi.Open("Show and Hide Demo", 1024, 640, false, World)
+	opts := engi.RunOptions{
+		Title:  "Show and Hide Demo",
```
matiwinnetou: :+1:

demos/hud/hud.go
```diff
@@ -100,5 +100,10 @@ func (*Game) Show()        {}
 func (*Game) Type() string { return "Game" }
 
 func main() {
-	engi.Open("HUD Demo", 400, 400, false, &Game{})
+	opts := engi.RunOptions{
```
matiwinnetou: so how can we make a full screen now? The flag was removed?

demos/pong/pong.go
```diff
@@ -285,6 +285,11 @@ func (ScoreMessage) Type() string {
 }
 
 func main() {
-	engi.SetScaleOnResize(true)
-	engi.Open("Pong", 800, 800, false, &PongGame{})
+	opts := engi.RunOptions{
```
matiwinnetou: much better, immutable approach.

demos/hide/hide.go
EtienneBruines: Also has to do with the OpenGL buffers. Because for most things, the Scale / Drawable doesn't change every frame, we're only re-computing these stuff on-change, and putting it in the GPU RAM (instead of regular RAM), which is way faster.

ecs/entity.go
```diff
@@ -8,7 +8,7 @@ import (
 // Systems, and has a number of Components
 type Entity struct {
 	id         string
-	components map[string]Component
+	components []Component
```
matiwinnetou: is this faster than a map?

demos/hud/hud.go
EtienneBruines:
```go
opts := engi.RunOptions{
    Title: "HUD Demo",
    Fullscreen: true,
}
```

engi.go
```diff
@@ -25,32 +21,53 @@ var (
 	scaleOnResize   = false
 	fpsLimit        = 120
 	headless        = false
+	vsync           = true
```
matiwinnetou: I never understood the idea of vsync.

EtienneBruines: VSync refers to the refresh-rate of the monitor. A game can update the graphics (for example) 120 times per second. But if the monitor only refreshes 60 times per second, half of those graphics-updates aren't that useful - most of the time, you won't see them anyway. 

VSync tells the GPU to "synchronize" with the monitor: make sure we're only doing GPU rendering when the monitor is refreshing. This usually means, in practice, the FPS-limit becomes `max(user_defined_fpsLimit, monitor_refresh_rate)`, in pseudo-code.

engi.go
```diff
@@ -25,32 +21,53 @@ var (
...
+	HeadlessMode bool
 
-	headless = true
+	Fullscreen    bool
+	Width, Height int
 
-	runHeadless(defaultScene)
-}
+	// VSync indicates whether or not OpenGL should wait for the monitor to swp the buffers
+	VSync bool
```
EtienneBruines: Should be changed to `DisableVSync`, because this defaults to `false`, and the default behavior should be to honor vsync.

demos/hud/hud.go
matiwinnetou: :+1:

engi.go
matiwinnetou: thanks!

collision.go
```diff
@@ -66,7 +63,7 @@ func (cs *CollisionSystem) New(*ecs.World) {
 }
 
 func (cs *CollisionSystem) RunInParallel() bool {
```
EtienneBruines: Not sure if this `RunInParallel` function is still being used.

demos/headless/headless.go
```diff
@@ -6,9 +6,10 @@ import (
 	"math/rand"
 	"sync"
```
EtienneBruines: TODO: remove newline between the groups of imports

ecs/entity.go
EtienneBruines: Usually, it is. I forgot if it actually made A LOT of performance improvement, or just a little...

ecs/system.go
```diff
@@ -57,27 +56,24 @@ func (s System) New()  {}
...
 func (s *System) AddEntity(entity *Entity) {
-	s.EntityMap[entity.ID()] = entity
+	s.entities = append(s.entities, entity)
 }
 
 func (s *System) RemoveEntity(entity *Entity) {
-	delete(s.EntityMap, entity.ID())
+	for entityIndex := range s.entities {
+		if s.entities[entityIndex].id == entity.id {
+			// TODO: delete it, then return
```
EtienneBruines: TODO :fearful:

engi_glfw.go
```diff
@@ -144,8 +137,8 @@ func run(defaultScene Scene, title string, width, height int, fullscreen bool) {
 				}
 
 				for _, s := range scene.world.Systems() {
-					if render, ok := s.(*RenderSystem); ok {
-						render.defaultBatch.SetProjection(gameWidth, gameHeight)
+					if _, ok := s.(*RenderSystem); ok {
+						Shaders.def.SetProjection(gameWidth, gameHeight)
 					}
```
EtienneBruines: TODO: we shouldn't call `SetProjection` to those specific shaders..., at least not within a loop which just checks 'do we have a RenderSystem'.

engi_glfw.go
```diff
@@ -170,12 +164,18 @@ func run(defaultScene Scene, title string, width, height int, fullscreen bool) {
...
 
-	runLoop(defaultScene, false)
+func DestroyWindow() {
+	glfw.Terminate()
 }
 
 func SetTitle(title string) {
-	window.SetTitle(title)
+	if headless {
+		log.Println("Title set to:", title)
```
EtienneBruines: TODO; we perhaps shouldn't log this?

render.go
EtienneBruines: TODO is resolved.

render.go
```diff
@@ -49,32 +41,158 @@ func (renderChangeMessage) Type() string {
...
+	var x3 float32
+	var y3 float32
+	var x4 float32
+	var y4 float32
+
+	if rotation != 0 {
+		rot := float64(rotation * (math.Pi / 180.0))
+
+		cos := float32(math.Cos(rot))
+		sin := float32(math.Sin(rot))
```
EtienneBruines: TODO: use `luxengine/math`

collision.go
EtienneBruines: It is, but not in the new PR. Leaving it as-is .

ecs/entity.go
EtienneBruines: Reverted; doesn't fit in this PR.

EtienneBruines:
Processed all TODO's and comments that I feel should/could be part of this PR. 

Merge when you feel you're ready for it, @paked.

engi_glfw.go
paked: If in doubt: Don't log

paked:
Wooo @EtienneBruines :heart:

PR #138, opened 2015-12-28
Changed Systemer to System interface

### LINKED ISSUE #119: Systems without Entities
Maybe it doesn't completely capture the ECS paradigm, but it has happened to me more than once. 

Whenever the `System` has to do something with time (spawn creep after 5 seconds), or something with keys (show menu on ESC), I have to add a dummy `Entity` for this. If it doesn't have any `Entity`s, it's never called.

Would it be OK to call `Pre()` and `Post()` regardless of the amount of `Entity`s? @paked

EtienneBruines:
Fixes #119

# Breaking change
Things like
```go
type MyCustomSystem struct {
    *ecs.System
}
```
have become
```go
type MyCustomSystem struct {
    *ecs.BasicSystemWrapper
}
```
----------------
Things like
```go
func (m *MyCustomSystem) New(*ecs.World) {
    m.System = ecs.NewSystem()
}
```
have become
```go
func (m *MyCustomSystem) New(*ecs.World) {
    m.BasicSystemWrapper = ecs.NewBasicSystemWrapper(m)
}
```
------------
Things like
```go
func (*MyCustomSystem) Update(*ecs.Entity, float32) {}
```
have become
```go
func (*MyCustomSystem) EUpdate(*ecs.Entity, float32) {}
```

# New features
A valid `System` (which you can add to a `World` instance), is this:
```go
type MyCustomSystem struct {}
func (*MyCustomSystem) Type() string { return "MyCustomSystem" }
func (*MyCustomSystem) Priority() int { return 1 }
func (*MyCustomSystem) AddEntity(*ecs.Entity) {}
func (*MyCustomSystem) RemoveEntity(*ecs.Entity) {}
func (*MyCustomSystem) New(*ecs.World) {}
func (*MyCustomSystem) Update(dt float32) {}
```
One can still use the 'old' behavior which does the looping-over-entities for you, this can be done like:
```go
type MyCustomSys struct {
    *ecs.BasicSystemWrapper
}
func (*MyCustomSys) Type() string { return "MyCustomSys" }
func (*MyCustomSys) Pre() {}
func (*MyCustomSys) Post() {}
func (m *MyCustomSys) New(*ecs.World) {
    m.BasicSystemWrapper = ecs.NewBasicSystemWrapper(m)
}
func (*MyCustomSys) EUpdate(*ecs.Entity, float32) {}
```

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -71,28 +71,26 @@ func (*AnimationComponent) Type() string {
 }
 
 type AnimationSystem struct {
-	*ecs.System
+	ecs.LinearSystem
 }
 
-func (a *AnimationSystem) New(*ecs.World) {
-	a.System = ecs.NewSystem()
-}
+func (a *AnimationSystem) New(*ecs.World) {}
 
-func (AnimationSystem) Type() string {
-	return "AnimationSystem"
-}
+func (*AnimationSystem) Type() string { return "AnimationSystem" }
+func (*AnimationSystem) Pre()         {}
+func (*AnimationSystem) Post()        {}
 
-func (a *AnimationSystem) Update(e *ecs.Entity, dt float32) {
+func (a *AnimationSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 	var (
 		ac *AnimationComponent
 		r  *RenderComponent
 		ok bool
 	)
 
-	if ac, ok = e.ComponentFast(ac).(*AnimationComponent); !ok {
+	if ac, ok = entity.ComponentFast(ac).(*AnimationComponent); !ok {
 		return
 	}
-	if r, ok = e.ComponentFast(r).(*RenderComponent); !ok {
+	if r, ok = entity.ComponentFast(r).(*RenderComponent); !ok {
 		return
 	}
 
--- a/demos/animation/anim.go
+++ b/demos/animation/anim.go
@@ -65,18 +65,16 @@ func (game *GameWorld) CreateEntity(point *engi.Point, spriteSheet *engi.Sprites
 }
 
 type ControlSystem struct {
-	*ecs.System
+	ecs.LinearSystem
 }
 
-func (ControlSystem) Type() string {
-	return "ControlSystem"
-}
+func (*ControlSystem) Type() string { return "ControlSystem" }
+func (*ControlSystem) Pre()         {}
+func (*ControlSystem) Post()        {}
 
-func (c *ControlSystem) New(*ecs.World) {
-	c.System = ecs.NewSystem()
-}
+func (c *ControlSystem) New(*ecs.World) {}
 
-func (c *ControlSystem) Update(entity *ecs.Entity, dt float32) {
+func (c *ControlSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 	var a *engi.AnimationComponent
 
 	if !entity.Component(&a) {
--- a/demos/animation/anim.go
+++ b/demos/animation/anim.go
@@ -36,7 +36,7 @@ func (game *GameWorld) Setup(w *ecs.World) {
 	w.AddSystem(&engi.RenderSystem{})
 	w.AddSystem(&engi.AnimationSystem{})
 	w.AddSystem(&ControlSystem{})
-	w.AddSystem(engi.NewMouseZoomer(zoomSpeed))
+	w.AddSystem(&engi.MouseZoomer{zoomSpeed})
 
 	spriteSheet := engi.NewSpritesheetFromFile("hero.png", 150, 150)
 
--- a/demos/falling/falling.go
+++ b/demos/falling/falling.go
@@ -117,7 +117,7 @@ func (rock *RockSpawnSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 }
 
 func NewRock(position engi.Point) *ecs.Entity {
-	rock := ecs.NewEntity([]string{"RenderSystem", "FallingSystem", "CollisionSystem", "SpeedSystem"})
+	rock := ecs.NewEntity([]string{"RenderSystem", "FallingSystem", "CollisionSystem"})
 
 	texture := engi.Files.Image("rock.png")
 	render := engi.NewRenderComponent(texture, engi.Point{4, 4}, "rock")
@@ -136,12 +136,8 @@ type FallingSystem struct {
 }
 
 func (*FallingSystem) Type() string { return "FallingSystem" }
-func (*FallingSystem) Pre()         {}
-func (*FallingSystem) Post()        {}
 
-func (fs *FallingSystem) New(*ecs.World) {
-	//engi.Mailbox.Listen("CollisionMessage", fs)
-}
+func (fs *FallingSystem) New(*ecs.World) {}
 
 func (fs *FallingSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 	var space *engi.SpaceComponent
@@ -156,8 +152,6 @@ type DeathSystem struct {
 }
 
 func (*DeathSystem) Type() string { return "DeathSystem" }
-func (*DeathSystem) Pre()         {}
-func (*DeathSystem) Post()        {}
 
 func (ds *DeathSystem) New(*ecs.World) {
 	// Subscribe to ScoreMessage
@@ -170,9 +164,7 @@ func (ds *DeathSystem) New(*ecs.World) {
 	})
 }
 
-func (fs *DeathSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
-
-}
+func (fs *DeathSystem) UpdateEntity(entity *ecs.Entity, dt float32) {}
 
 func (fs *DeathSystem) Receive(message engi.Message) {
 	collision, isCollision := message.(engi.CollisionMessage)
--- a/demos/falling/falling.go
+++ b/demos/falling/falling.go
@@ -53,18 +53,16 @@ func (*Game) Show()        {}
 func (*Game) Type() string { return "Game" }
 
 type ControlSystem struct {
-	*ecs.System
+	ecs.LinearSystem
 }
 
-func (control *ControlSystem) New(*ecs.World) {
-	control.System = ecs.NewSystem()
-}
+func (*ControlSystem) Type() string { return "ControlSystem" }
+func (*ControlSystem) Pre()         {}
+func (*ControlSystem) Post()        {}
 
-func (ControlSystem) Type() string {
-	return "ControlSystem"
-}
+func (control *ControlSystem) New(*ecs.World) {}
 
-func (control *ControlSystem) Update(entity *ecs.Entity, dt float32) {
+func (control *ControlSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 	var space *engi.SpaceComponent
 	if !entity.Component(&space) {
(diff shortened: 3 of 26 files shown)


ecs/basicsystem.go
```diff
@@ -0,0 +1,79 @@
+package ecs
+
+import "runtime"
+
+// BasicSystemWrapper is the default implementation of the System interface.
+type BasicSystemWrapper struct {
```
matiwinnetou: maybe this wrapper is not necessary. In Java such classes would be named Adapter but some people call them Stub as well.

SystemStub or SystemAdapter.

ecs/world.go
```diff
@@ -61,7 +45,7 @@ func (w *World) RemoveEntity(entity *Entity) {
 }
 
 // AddSystem adds a new System to the World, and then sorts them based on Priority
-func (w *World) AddSystem(system Systemer) {
```
matiwinnetou: System is much better than Systemer to be honest :+1:

render.go
```diff
@@ -185,8 +184,8 @@ func (*RenderSystem) Type() string {
 	return "RenderSystem"
 }
 
-func (rs *RenderSystem) Priority() int {
-	return 1
+func (*RenderSystem) Priority() int {
+	return -10000 // some low value
```
matiwinnetou: perhaps it is better to use some constants, like Integer.MIN_VAL, does something like this exists in golang?

demos/falling/falling.go
```diff
@@ -91,21 +91,21 @@ func (control *ControlSystem) Update(entity *ecs.Entity, dt float32) {
 }
 
 type RockSpawnSystem struct {
-	*ecs.System
+	*ecs.BasicSystemWrapper
```
matiwinnetou: on a less dev friendly side, there is much more "typing" now.

demos/animation/anim.go
```diff
@@ -61,18 +61,18 @@ func (game *GameWorld) CreateEntity(point *engi.Point, spriteSheet *engi.Sprites
...
+func (*ControlSystem) Pre()         {}
+func (*ControlSystem) Post()        {}
 
 func (c *ControlSystem) New(*ecs.World) {
-	c.System = ecs.NewSystem()
+	c.BasicSystemWrapper = ecs.NewBasicSystemWrapper(c)
 }
 
-func (c *ControlSystem) Update(entity *ecs.Entity, dt float32) {
+func (c *ControlSystem) EUpdate(entity *ecs.Entity, dt float32) {
```
matiwinnetou: It may not be obvious what eupdate means

ecs/basicsystem.go
EtienneBruines: Well, the name doesn't really matter to me (perhaps @paked has a preference). But it *is* necessary: if I define these functions in some kind of struct, then there is no way to override the `Pre()`, `EUpdate()` and `Post()` functions.

ecs/world.go
EtienneBruines: I like having both; for a lot of things, the (currently) `BasicSystem` is also kind of nice. :smile:

render.go
EtienneBruines: Well, I don't want to use the minimum value, just a value that's low enough that people can use the values between this one and the default one (`0` or `1` or so). 

I could use 
```go
const RenderSystemPriority = -10000
```
but I don't know if that would add anything? What do you think, @matiwinnetou @paked?

ecs/basicsystem.go
paked: @EtienneBruines You could embed it within another struct?

demos/animation/anim.go
paked: Yeah, I don't like this function name.

Also: What is the difference here? The function signatures look identical, I don't think people could tell just be looking.

EtienneBruines: The difference is that the Update function takes only a time difference float as parameter, while EUpdate (E for Entity) takes both an Entity and the float. 

On December 31, 2015 11:15:30 PM GMT+01:00, Harrison Shoebridge <notifications@github.com> wrote:
>>  }
>>  
>> -func (c *ControlSystem) Update(entity *ecs.Entity, dt float32) {
>> +func (c *ControlSystem) EUpdate(entity *ecs.Entity, dt float32) {
>
>Yeah, I don't like this function name.
>
>Also: What is the difference here? The function signatures look
>identical, I don't think people could tell just be looking.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/paked/engi/pull/138/files#r48672646

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

demos/falling/falling.go
EtienneBruines: I could rename `BasicSystemWrapper` to `BasicSystem`, and rename `BasicSystem` to `BasicSystemFunctions`? (since the latter will probably never be used, only implemented)

@paked ?

ecs/basicsystem.go
EtienneBruines: Well, if it's in a struct, then there's no way to 'override' those methods. This problem doesn't exist for interfaces.

demos/animation/anim.go
paked: Ah. OK. I See now.

render.go
matiwinnetou: It may be better, otherwise one introduces magic values...

paked: Sure.

matiwinnetou:
Why is this abandoned?

EtienneBruines:
Haven't had the time, lately... Should have more time after 2nd of February. Feel free to work on it if you want, though.

paked:
@faide What are your thoughts on this PR?

I'm very wary it is a drastic breaking change, but I think will be worth it in the long run. Perhaps this can be one of our last breaking changes on the road to stability.

faide:
I clearly see the point of having "systems without entities"

Let me explain what I understood:

we will have a System interface like so:

```go
type System interface {
    Update(dt float32)
}

type EntitySystem interface {
    EntityUpdate(entity *Entity, dt float32)
}
```

The new way engi will work is by calling Update() on every registered "System interface". The system will be responsible to select the entities (if any) it wants to manage

If one desires to still have a system called by engi for each relevant entity she should "inherit from the SystemStub by including a pointer to it in its system struct and then implement the EntitySystem to be called per entity instead of globally.

But since that world.AddSystem() only takes a System interface it will only see the update() method...

I suspect the StubSystem will indeed be an implementer of the System interface and in this special implementation will try to type assert itself into an EntitySystem interface in order to be able to call EntityUpdate() for each relevant entity.

I like the idea. I feel we should take some time to break the API while we still can.

paked:
@EtienneBruines Thoughts on above?

EtienneBruines:
I think what @faide has understood is correct. The `World` will only take a `System` interface with the `Update(dt float32)` method. The `StubSystem` (or `BasicSystemWrapper` as used in the code), implements this `Update(dt float32)` function, by iterating over all entities and calling the user-defined `Update(entity *Entity, dt float32)` function.

animation.go
```diff
@@ -61,28 +62,28 @@ func (*AnimationComponent) Type() string {
 }
 
 type AnimationSystem struct {
-	*ecs.System
+	*ecs.BasicSystemWrapper
```
EtienneBruines: Should rename to `LinearSystem`.

EtienneBruines:
New commit includes:
`EUpdate` is renamed to `UpdateEntity`
`BasicSystemWrapper` is renamed to `LinearSystem`
`NewBasicSystemWrapper` is no longer needed. 

All references to `*BasicSystemWrapper` are now changed to `LinearSystem` (note that it's no longer a pointer-reference. 

TODO:
* `BasicSystem` is kinda a weird name, so thinking if we should rename it. 
* Logging a warning whenever someone *does* use a pointer-reference.

EtienneBruines:
Merged all commits into one. Could either of you review the code? @faide @paked? 

The reason behind all of this, is to allow for `Update(dt float32)` (so without having to define `Pre`, `Post` and Update-per-entity), and also allow for `Update(e *ecs.Entity, dt float32)` for those that want it.

paked:
@EtienneBruines I'm happy to review this. Preliminary checks says it looks OK, but it is almost 2am... So I won't merge this until awake me has approved it xD.

Thanks for your hard work on this.

EtienneBruines:
I do want to note that there are probably some systems that can be changed now (they're a `LinearSystem`, but don't benefit from `UpdateEntity` over `Update`), but that could be done in another stage/commit.

EtienneBruines:
I just thought of something. Shouldn't the `Pre()` and `Post()` methods of a `LinearSystem` require a `dt float32` parameter? (for the time in seconds since last frame?)

It's not a must, but I can imagine this being useful - since there's no way to know the `dt` inside those `Pre()` functions (theoretically, you can know it for the `Post` if you saved it in `UpdateEntity`, but that's ugly). 

Thoughts?

faide:
I have merged and tested this and my project builds and runs... But I have a gripe with the current implementation: during conversion I had some linear systems and I forgot to implement the LinearSystem interface (forgot to implement an empty Pre() method).

The compiler did not catch it (could not) and I got a run time panic at the moment I setup the scene that added this system (not before).

This means in a big project it will be quite easy to induce runtime panics by error.

To solve this I would prefer another method on World like : AddLinearSystem(s LinearSystem)
this one would wrap the normal add system and do the type cast. This would allow compile time detection of the issue instead of a runtime panic which is way nicer imho.

Regarding the dt being passed to Pre() & Post() it seems quite reasonable.

paked:
@faide I don't quite see how this panic could occur.

Would you mind posting the `struct` definition of your system?

In our examples we used to embed `ecs.System` as a pointer, which required the initialisation `as.System = &ecs.System{}`. However, with `ecs.LinearSystem` you no longer need to embed with the pointer. I have a feeling this is the mistake.

paked:
I actually purposely did not include `dt` in `Pre` or `Post` as a design decision. They are intended to be used for startup/cooldown operations (Like setting up a Batch or caching assets), instead of actually interacting with the game world.

If there is enough demand, I suppose we could put it in.

I just feel that if we expose `dt` everywhere, the new `Update` function will become meaningless.

EtienneBruines:
@faide @paked The panic occurs whenever we try to cast the system we're adding to `ecs.BasicSystem` (if it's a `ecs.LinearSystem`). I suppose I can either wrap it, and make the `Pre()` and `Post()` functions optional. Should we?

The error message was meant to help debug the problem, which in this case, did its job.

paked:
@EtienneBruines I'd prefer wrapping it.

Lets get these merge conflicts solved, and then we can get this merged in!

EtienneBruines:
Wrapped it nicely, `Pre` and `Post` are optional, and you can have one without the other. 

The `UpdateEntity` is mandatory (on purpose). If you're not implementing it, you shouldn't use `LinearSystem` to begin with. 

Also updated all demos to remove all optional (empty) `Pre` and `Post` functions. Some of them might even benefit from changing from `LinearSystem` to the general `ecs.System`, but I think it's OK for now.

paked:
Alright. I'm going to merge this in and break *all* the code for everyone using us as a dependency!

paked:
Woo! Longest running PR ever has been closed.

Thanks for being persistent on this @EtienneBruines.

PR #139, opened 2015-12-29
minor tweaks

ghost:


input.go
```diff
@@ -4,9 +4,9 @@
 
 package engi
 
-type Action int
-type Key int
-type Modifier int
+type Action 	int
+type Key 	int
+type Modifier 	int
```
EtienneBruines: Is this according to `gofmt` or `goimports`? I feel like this will be reverted the next time we're running `gofmt` (which is all the time).

font.go
```diff
@@ -6,8 +6,8 @@ import (
 	"image/draw"
 	"io/ioutil"
 	"log"
-
 	"fmt"
+
```
EtienneBruines: Good catch.

input.go
paked: Yeah, I've never seen this before.

EtienneBruines: I have seen it, I think ...  in my IDE (`WebStorm` / `IDEA` (with [Go plugin](https://github.com/go-lang-plugin-org/go-lang-idea-plugin)) - but this gets reverted when I run `gofmt`.

paked:
Sorry @ghost. I'm trying to clean up our PRs at the moment, this looks like it is being submitted in a parallel PR already.

PR #146, opened 2016-03-01
Handle Mouse button differentiation, handle button release & only send one event

faide:


paked:
@faide, my laptop just died. I might take a bit to review this.

EtienneBruines:
Nice PR :smile:. I have reviewed all commits, and commented on the code. 

Feel free to ask any questions you may have, and I will do my best to answer them.

collision.go
```diff
@@ -16,6 +16,16 @@ type SpaceComponent struct {
 	Height   float32
 }
 
+// Center positions the space component according to its center instead of its
+// top-left point (this avoids doing the same math each time in your systems)
+func (sc *SpaceComponent) Center(p Point) {
```
paked: Hmm. I'm not sure this is the best way to approach this problem.

For example:

**Sprite A**:

- Position: `0, 0`
- System: `DoesNotCallCenter`
- Width: `20`


**Sprite B**:

- Position: `0, 0`
- System: `CallsCenter`
- Width: `20`

If this code was run, **Sprite B** would end up at position `(10, 10)`, while **Sprite A** kept the same position - Even if the system made no other changes.

input.go
```diff
@@ -10,12 +10,35 @@ var (
...
+	MouseButton8      MouseButton = 7
+	MouseButtonLast   MouseButton = 7
+	MouseButtonLeft   MouseButton = 1
+	MouseButtonRight  MouseButton = 2
+	MouseButtonMiddle MouseButton = 3
+)
+
+// these values are remapped during engin_glfw init()
+// so I suppose these are the default mappings for engi_js
+// TODO: can someone confirm this by removing this comment and setting a proper one ?
```
paked: Yes. It appears so. Could you move them to `engi_js.go`?

paked:
@faide I've given your code a review and left some small comments. Once we have those resolved, I'll be happy to merge it in.

Thanks for your persistence!

input.go
faide: sure. Will do.

collision.go
faide: Is it the name that you don't like? Maybe we should call it PositionCenter
The whole point of this method is to be able to move an entity by placing "it's center" to a desired point, but the name does not fully reflect that. An since this is just a helper I can place it anywhere in my own code but I felt like it could be interesting to reuse.

input.go
faide: After looking a bit more into this code I think we won't move that so easily. engi_js.go is not always used as part of the build but those are global variables used by both engi_js and engi_glfw ...

paked: Could we define them as

````go
var (
Dash Key
Apostrophe Key
)
```

in this file, and then give them values in `engi_js.go` and `engi_glfw.go`?

faide:
Paked I agree we could move the value affectation into each file (js & glfw) but this would create many duplicates because only a small subset is redefined in glfw... Would this be acceptable?

paked:
@faide Ah ha. Now I see the problem. OK, I'd leave it as it is then.

Make any final changes, then I'm happy to merge this in.

faide:
I think I finished what I could in this PR, let's see what will be the next one :)

paked:
@faide Thanks for the PR. It is awesome to see other people contributing!

PR #148, opened 2016-03-07
Detect unmet dependencies in entities at adding time to world

faide:
Hi,

Here is the beginning of a PR / Proposition to add some control / error checking on the entity requirements. ATM the PR only adds logging to warn the developer an entity is being added to a world that does not have the sufficient systems to support its operation.

We could add an error return to the  (*ecs.World).AddEntity signature and this would break all previous code relying on the old signature. Should we make this breaking change while we can or should we just keep the logging version?

Introducing this new feature I added a new private field (systemMap) to ecs.World for my own use and I also took advantage of it in (*ecs.World).HasSystem() implementation.

### DIFF
--- a/ecs/entity.go
+++ b/ecs/entity.go
@@ -32,6 +32,18 @@ func (e *Entity) DoesRequire(name string) bool {
 	return e.requires[name]
 }
 
+// Requirements returns a list of all system requirements for this entity
+func (e *Entity) Requirements() []string {
+	keys := make([]string, len(e.requires))
+
+	i := 0
+	for k := range e.requires {
+		keys[i] = k
+		i++
+	}
+	return keys
+}
+
 // AddComponent adds a new Component to the Entity
 func (e *Entity) AddComponent(component Component) {
 	e.components[component.Type()] = component
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -1,15 +1,16 @@
 package engi
 
 import (
-	"github.com/paked/engi/ecs"
 	"testing"
+
+	"github.com/paked/engi/ecs"
 )
 
 type NilSystem struct {
 	*ecs.System
 }
 
-func (ns *NilSystem) New() {
+func (ns *NilSystem) New(*ecs.World) {
 	ns.System = ecs.NewSystem()
 }
 
--- a/camera.go
+++ b/camera.go
@@ -87,7 +87,7 @@ func (cam *cameraSystem) moveToX(location float32) {
 }
 
 func (cam *cameraSystem) moveToY(location float32) {
-	cam.y = mgl32.Clamp(location, WorldBounds.Min.X, WorldBounds.Max.Y)
+	cam.y = mgl32.Clamp(location, WorldBounds.Min.Y, WorldBounds.Max.Y)
 }
 
 func (cam *cameraSystem) zoomTo(zoomLevel float32) {
--- a/camera_test.go
+++ b/camera_test.go
@@ -3,11 +3,15 @@ package engi
 import (
 	"github.com/stretchr/testify/assert"
 	"testing"
+
+	"github.com/paked/engi/ecs"
 )
 
 func TestCameraMoveX(t *testing.T) {
+	w := &ecs.World{}
+	w.New()
 	cam := &cameraSystem{}
-	cam.New()
+	cam.New(w)
 	WorldBounds = AABB{Point{0, 0}, Point{300, 300}}
 
 	currentX := cam.X()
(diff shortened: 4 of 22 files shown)


EtienneBruines:
LGTM. I'm not sure which I like more:
- Letting `AddEntity` return an `error`, or
- Logging whenever a dependency is not met. 

I do have to note that, because `AddEntity` currently returns nothing, it won't be a breaking change if it returns an `error` from now on. 
This is because this usage is legal in both cases:
```go
func example() {
    e := ecs.NewEntity(nil)
    w.AddEntity(e)  // because this line does not have an assignment, ever, it's not a breaking change
}
```

faide:
Yep you're right it does not break existing code :) The question now is should we return an error or just log things off and be done?

In the way I see general programming this should return and error handled or not by the caller (not the function)

In my eyes if I add a system requirement on an entity and the world I am adding this entity to does not support my requirements I should receive an error and be able to do something. (Be it abort, or log to the user display during beta builds... or whatever I intend to do).

The counter argument to this would be: you need to test your game and the library does not have to burden all users with coding-safety performance impacts to help you during the dev phase...

But all in all I prefer to add some coding safety and help build my game(s) faster that being a low level ultra optimized 2d engine that makes my debugging sessions harder and force me to play catch and mouse at run time because the low coupling dependency declaration does not check anything.

The real question though... Is is conceivable that someone adds an entity to an "unfinished" world and add the required systems to its world later on? In this use case an error return may not be optimum.

EtienneBruines:
I can imagine one adding a dependency on the entity (i.e. depending on `MouseSystem`), but then forgetting to add the `MouseSystem`, because it wasn't needed until that moment. So that would be an acceptable use-case.

faide:
Seeing the way the (*ecs.World).AddEntity() is implemented (ie: adding the entity to the relevant systems existing at the time of callit it) I suspect it is intended in the API usage that the developer first adds all the desired systems to the world then and only then begins adding entities to the said world.

This is even more in favor of error checking and error returning at the AddEntity() moment.

@paked ? Any idea on the question? To return error or to not return error, that is the question :)

ecs/entity.go
```diff
@@ -32,6 +32,18 @@ func (e *Entity) DoesRequire(name string) bool {
 	return e.requires[name]
 }
 
+// Requirements returns a list of all system requirements for this entity
```
paked: Any reason why you did not just return `e.requires`?

paked:
@EtienneBruines @faide Thoughts on panicking? I can not think of a single situation where it actually makes sense for a developer to add an `Entity` which requires a system which has not been added to the `World` already. Given the fact 1.6 added *much* **much** nicer panic logs, we could probably get behind this?

EtienneBruines:
@paked The downside with panicking, is that there might be a valid reason for not having added those `System`s yet - such as not-yet-implemented.

ecs/entity.go
EtienneBruines: Because that can panic whenever the key does not exist within the map.

faide:
I am not in favour of panicking in this kind of situation. We are a library and should avoid panic situations as much as possible. The caller is responsible of handling any error if we choose to return an error.
The issue with panic is that it will happen at runtime and only when (and if) you add the specific entity that requires the missing system. And runtime crashes are really bad in any program and even more in games. If we look at the industry we see a lot of unfinished games with obvious bugs shipped to production. If such bugs led to runtime panic this would be even worse.

Now we can argue knowingly shipping bugs to production is bad, but shipping crashing programs to production is even worse... In our case forgetting a system will lead to missing comportments in your game and thus somewhat "visible" bugs... but at least it won't crash (instantaneously).

And as Etienne said I can imagine situations where I load entities from a data-table (cvs or whatever rulebook I choose) and the gameplay designer (a non coding person) can edit this rulebook and reference not-yet-added systems we just discussed together about. In this case I prefer to handle missing system errors gracefully as a way to remind me which system I need to implement and plug in my world rather than be forced to rush add a stub system to make my game testable by anyone in the team again.

paked:
@faide @EtienneBruines OK. Returning an error sounds satisfactory then. I don't think we need to log by default.

Perhaps we should make an "internal `engi` logger", which by default will not be used but can be enabled with something like:

```go
// in the end-users code
engi.LogTo = os.Stdout
```

Or something similar. (I think there is a logger interface that we could use as well.)

I think this would be really handy for debugging, but would allow our users to not have to deal with what they may see as unnecessary information.

ecs/entity.go
paked: OK. Fair enough.

EtienneBruines:
I like the idea of that.

faide:
OK Will do this tonight
Le 10 mars 2016 12:44 PM, "Etienne Bruines" <notifications@github.com> a
écrit :

faide:
Warning I merged with another fix branch I have. This adds modifier propagation to mouse Components and fixes a clipping bug in the camera.
This also begins to fix unit tests. Unfortunately tests are not yet rening because I have a panic in the camera test that I still don't understand correctly

paked:
Thanks for the PR @faide!

PR #152, opened 2016-03-08
Fixed issue where MouseComponent would stay Clicked

EtienneBruines:
Instead of resetting most values to false, we're now
forgetting every value from the last frame (with some
exceptions), to ensure things like `Click` don't fire twice. (which they do now)

Would you like to review @faide ?

### DIFF
--- a/mouse.go
+++ b/mouse.go
@@ -87,6 +87,9 @@ func (m *MouseSystem) Pre() {
 
 // Update sets the MouseComponent values for each Entity
 func (m *MouseSystem) Update(entity *ecs.Entity, dt float32) {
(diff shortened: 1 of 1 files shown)


EtienneBruines:
@faide I have pushed a new version, which pushes the functionality of `Track` a bit upwards (now it works without having a `SpaceComponent`). 

Is this still in line with your thoughts of `Track`, or did I misinterpret? 

Feel free to criticize.

paked:
@EtienneBruines I'm happy for you to merge these this in once you have resolved @faide's comment :)

faide:
@EtienneBruines could you merge this one?

PR #153, opened 2016-03-08
Font.TTF is now public

EtienneBruines:
`font.TTF` is now public. Even though it's mostly used internally,
it's nice to be able to read/change the value from the outside.

I also refactored `RenderNRGBA` from the `Render` function - this is because someone might not want a `Texture`, but only an intermediary `NRGBA` (possibly doing some stuff to it, before creating a `Texture` from it).

### DIFF
--- a/font.go
+++ b/font.go
@@ -1,13 +1,13 @@
 package engi
 
 import (
+	"fmt"
 	"image"
 	"image/color"
 	"image/draw"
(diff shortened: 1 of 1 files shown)


paked:
Thanks @EtienneBruines. Sorry it took a while to get back to you on this.

PR #155, opened 2016-03-16
Rotation support

faide:
Hi there,

this is a PR (based on other ones) that adds rotation support for Engi.

The rotation is implemented in the default shader. I'd like the changes I introduced in the shader interface.It may be possible to avoid this by passing the space width and height in the gl buffer.

### DIFF
--- a/collision.go
+++ b/collision.go
@@ -136,7 +136,6 @@ func (cs *CollisionSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 				}
 
 				Mailbox.Dispatch(CollisionMessage{Entity: entity, To: other})
-				collision.Collides = true
 				collided = true
 			}
 		}
--- a/collision.go
+++ b/collision.go
@@ -14,6 +14,7 @@ type SpaceComponent struct {
 	Position Point
 	Width    float32
 	Height   float32
+	Rotation float32 // angle in degrees for the rotation to apply clockwise
 }
 
 // Center positions the space component according to its center instead of its
--- a/collision.go
+++ b/collision.go
@@ -15,7 +15,7 @@ type SpaceComponent struct {
 	Position Point
 	Width    float32
 	Height   float32
-	Rotation float64 // angle in degrees for the rotation to apply clockwise
+	Rotation float32 // angle in degrees for the rotation to apply clockwise
 }
 
 // Center positions the space component according to its center instead of its
--- a/collision.go
+++ b/collision.go
@@ -50,6 +50,7 @@ func (cm CollisionMasterComponent) Is() bool {
 type CollisionComponent struct {
 	Solid, Main bool
 	Extra       Point
+	Collides    bool // Collides is true if the component is colliding with something during this pass
 }
 
 func (*CollisionComponent) Type() string {
@@ -104,6 +105,7 @@ func (cs *CollisionSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 		otherCollision *CollisionComponent
 	)
 
+	collided := false
 	for _, other := range cs.Entities {
 		if other.ID() != entity.ID() {
 			if otherSpace, ok = other.ComponentFast(otherSpace).(*SpaceComponent); !ok {
@@ -134,9 +136,13 @@ func (cs *CollisionSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 				}
 
 				Mailbox.Dispatch(CollisionMessage{Entity: entity, To: other})
+				collision.Collides = true
+				collided = true
 			}
 		}
 	}
+	// if no one collided reset to false
+	collision.Collides = collided
 }
 
 func IsIntersecting(rect1 AABB, rect2 AABB) bool {
--- a/collision.go
+++ b/collision.go
@@ -14,7 +14,7 @@ type SpaceComponent struct {
 	Position Point
 	Width    float32
 	Height   float32
-	Rotation float32 // angle in degrees for the rotation to apply clockwise
+	Rotation float64 // angle in degrees for the rotation to apply clockwise
 }
 
 // Center positions the space component according to its center instead of its
--- a/render.go
+++ b/render.go
@@ -111,7 +111,6 @@ func (ren *RenderComponent) preloadTexture() {
 func (ren *RenderComponent) generateBufferContent() []float32 {
 	scaleX := ren.scale.X
 	scaleY := ren.scale.Y
-	rotation := float32(0.0)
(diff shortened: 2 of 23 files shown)


paked:
OK. I think you're going to need to update to master.

Before I merge this in, would you mind squashing the new commits?

faide:
@paked sure!

Clearly I don't expect you to merge this as is... :)
I want to spark a discussion on some design choices. And I need to polish quite a few things first.

Per example I'd like to give the user access to the rotation axis as a uniform.

I am not that happy with the way the projection is calculated at the moment and I'd like to add a projection matrix and remove the old camera + projection vars. But doing so will need to change quite a few things in the shader interface.

I would also like to let the user control fov and near plane / far plane distance (with sane defaults) to allow for easy paralax effects (just to give one exemple of usage).

EtienneBruines:
"I am not that happy with the way the projection is calculated at the moment" -> if you have any alternatives, I'd be open to hearing them :-). 

"I'd like to add a projection matrix" -> that sounds like a good thing to do

"remove the old camera + projection vars." -> I guess that's ok. 

I would also like to let the user control fov and near plane / far plane distance -> this is currently done using the Camera (by zooming, with some defaults) - but if you have another idea, sure. 

Going to look at your commits now.

paked:
:+1: this is so hype

paked:
@faide What is the status on this, ready to merge?

shaders.go
```diff
@@ -9,7 +12,7 @@ const bufferSize = 10000
 type Shader interface {
 	Initialize(width, height float32)
 	Pre()
-	Draw(texture *webgl.Texture, buffer *webgl.Buffer, x, y, rotation float32)
+	Draw(texture *webgl.Texture, buffer *webgl.Buffer, width, height, x, y float32, rotation float64)
```
EtienneBruines: Why not use `float32` for this? (as we're using `float32` everywhere else)

(I know native `float64` is faster in Go, but conversion between them is even worse; see tests at https://github.com/luxengine/math)

EtienneBruines:
Other than changing `float64` to `float32`, this LGTM.

faide:
This comes from the fact that I need to do : 
```go
Gl.Uniform1f(s.ufS, float32(math.Sin(angle)))
```

and math.Sin and math.Cos need float64 so I need a float64

If I explicitely use float32 this will become:: 
```go
Gl.Uniform1f(s.ufS, float32(math.Sin(float32(angle))))
```

which will be worse.

EtienneBruines:
You could use `github.com/luxengine/math` as your `math` library (instead of the builtin `math`). It's pretty much the same, but uses `float32`.

faide:
I still "feel" (no evidence) that doing the trigonometry calculations on the CPU once is better that doing it on the GPU once per vertex.

faide:
wow did not see luxengine post. Would this be ok to add a dependency? If yes then this would be the way to go!

faide:
I see that luxengine does not "care" about architecture other that i386 and amd64... would this be an issue for us ? mobile devices may have other arch...

EtienneBruines:
Well, we're already using `luxengine` at other parts of `engi`, so I think it'd be OK to use it for this.

paked:
@faide Awesome. You're going to need to fix these merge conflicts before we can merge it in :dancer:

paked:
@faide Do you think you could make a demo showcasing the rotation? If not, I'd be happy to do this in a follow up PR.

faide:
I can make a small demo.

paked:
@faide Awesome.

render.go
```diff
@@ -152,7 +151,7 @@ func (ren *RenderComponent) generateBufferContent() []float32 {
 	red := colorR
 	green := colorG << 8
 	blue := colorB << 16
-	alpha := uint32(transparency*255.0) << 24
+	alpha := uint32(ren.Transparency*255.0) << 24
```
EtienneBruines: This reminds me of an issue ... `ren.Transparency` should be private, and using a `set`-function. Not because we like it, but because the `generateBufferContent` has to be called again after changing it.

We could also (instead) rename `generateBufferContent` to `FlushCache`, and require the user to call it whenever he changed things like color, transparency, texture, scale. 

Thoughts @paked @faide?

collision.go
```diff
@@ -134,9 +136,13 @@ func (cs *CollisionSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 				}
 
 				Mailbox.Dispatch(CollisionMessage{Entity: entity, To: other})
+				collision.Collides = true
```
EtienneBruines: This line (139) isn't needed, as this is being done (again) in line 145.

faide: Yep seen that yesterday night after pushing. But I was too tired to cleanup. But I agree.

render.go
faide: Well in fact my idea on this (this is not final) is that I should be able to control transparency level from any system that wants to do so (as a visual feedback to the used, eg  "you cannot place the building here because of a collision ^^")

This leads me to the idea that the transparency level should be sent to the shader as a uniform and removed from the buffer altogether, and then used in the fragment shader to produce the final color during rendering.

If we do so Transparency should certainly stay public. If we don't do this we'll have to recompute the whole buffer again which seems like a waste to me because fragment shaders are here to compute colors for us anyway...

My concern with this is: will we degrade performance a lot doing so?

paked: Personally, I am fine with the current implementation - Regenerating the buffer seems alright to me. Realistically, I don't think it will be happening that often. If it becomes an issue in the future, changing it up should not be too much of an issue.

faide:
I think I am in a correct state for a merge. What do you @paked & @EtienneBruines think?

faide:
After merging from origin:master, I mean :p

faide:
Transparency does not have any visible effect (before or after my changes) we'll have to investigate more but I think it should be in another PR.

EtienneBruines:
The `Transparency` variable indeed has no effect at the moment (we'd have to debug that in another issue). 

LGTM.

paked:
@faide Would you mind bringing this back up to master so we can get it merged in ASAP?

faide:
Will definitely do. Since it was up to the master yesterday night I expect
the merge to be quite simple.
Le 2 avr. 2016 12:48 AM, "Harrison Shoebridge" <notifications@github.com> a
écrit :

paked:
@faide Do you want me to merge this in myself if you don't have time anymore?

EtienneBruines:
Closed. See https://github.com/EngoEngine/engo/pull/216.

PR #158, opened 2016-03-23
Add duplication to Entity

### LINKED ISSUE #104: Presets (prefabs?) for the easy re-initialization of an entity pattern
As the old `README.md` stated, we might want to:
```
Presets (prefabs?) for the easy re-initialization of an entity pattern
```

I'm not sure what's wrong with the current implementation, or what we'd want to change. 

If @paked could clarify, or close this issue, that'd be great.

paked:
Fix #104 

I think there may be a problem if the Entity has pointers to Components in its `e.components` map. I'm not quite sure what the best way to fix this is though.

### DIFF
--- a/ecs/entity.go
+++ b/ecs/entity.go
@@ -77,3 +77,14 @@ func (e *Entity) ComponentFast(c Component) interface{} {
 func (e *Entity) ID() string {
 	return e.id
 }
+
+// Duplicate creates an array of entities the size of amount, each with the same components and requirements. They will not be in the same systems
+func (e Entity) Duplicate(amount int) []Entity {
+	var entities []Entity
+
+	for i := 0; i < amount; i++ {
+		entities = append(entities, e)
+	}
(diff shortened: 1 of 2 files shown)


EtienneBruines:
Shouldn't the tests say if it works with pointer as well? :trollface: 

Could you add the following to the test-function (and see if it works)?
* Two components, one added with `&Component1{}` (pointer), and one with `Component2{}` (non-pointer). 
* Then duplicate as you do now
* Then change the value of both of those `Component`s, to something else -- for the original `Entity` only
* Verify the values stay the same for the duplicates. 

I'm curious to see as to when it actually changes, and when it doesn't. Ideally, it wouldn't change them at all; in practice, I'd expect some things to change as well.

EtienneBruines:
I'm going to close it, and ask you to refile it [here](https://github.com/EngoEngine/ecs).

PR #160, opened 2016-03-25
Check to ensure paddles don't go out of the screen

otraore:
Fixes https://github.com/paked/engi/issues/159


paked:
Thanks for the fix @otraore!

PR #181, opened 2016-04-06
Changed PriorityLevel to ZIndex

### LINKED ISSUE #151: PriorityLevel is limiting
Currenty, we have a few PriorityLevel values:

```go
const(
	// HighestGround is the highest PriorityLevel that will be rendered
	HighestGround PriorityLevel = 50
	// HUDGround is a PriorityLevel from which everything isn't being affected by the Camera
	HUDGround    PriorityLevel = 40
	Foreground   PriorityLevel = 30
	MiddleGround PriorityLevel = 20
	ScenicGround PriorityLevel = 10
	// Background is the lowest PriorityLevel that will be rendered
	Background PriorityLevel = 0
	// Hidden indicates that it should not be rendered by the RenderSystem
	Hidden PriorityLevel = -1
)
```

As a reminder: the usage of these is as follows: the higher the `PriorityLevel`, the later it will be rendered (so it'll be on on top). This is done by having 51 lists of `Entity`-references, and looping over each of them (see https://github.com/paked/engi/blob/master/render.go#L255-L288 ). One for each value between `Background` and `HighestGround`. 

However, it would be nice to have limitless values in between (like an ordering, instead of a grouping). 

Why not multiply those values by a million? Well, in the current workflow, this would create millions of `Entity`-reference lists, and we would have to loop over each and every one of them - including those that are empty. Using a map instead isn't desirable either, because looping over a `map` also is quite resource demanding (when doing it 50 million times within `6.944ms` (on 144hz)). 

----------------------

Possible fix: using `sort.Sort`, to actually sort based on the `PriorityLevel`. The values given at the top of this issue, can be scaled from anything between `0` and `2^32-` or so (`uint32`). When does this sorting have to take place? Well, all `Entity`s that are already part of the `RenderSystem`, are ordered. Whenever you add a new item to the `RenderSystem`, that last item would be unsorted. This gives us two possibilities:
- Sorted insertion (i.e.: we lookup where we should add it, and add it there (instead of at the end);
- Or: call sort.Sort in the `(r *RenderSystem) Post()` function, so it'll automatically sort in every added `Entity`. 

I could implement either one of those things, but before I do so, it'd be nice to hear your thoughts @paked @faide ?

EtienneBruines:
Should fix #151. Changed default FPS from 120 to 60. 

Changed resolution of `ZoomDemo` from 400x400 to 900x900. 

Please review.

### DIFF
--- a/camera.go
+++ b/camera.go
@@ -3,8 +3,8 @@ package engo
 import (
 	"sync"
 
-	"github.com/go-gl/mathgl/mgl32"
 	"github.com/engoengine/ecs"
+	"github.com/go-gl/mathgl/mgl32"
 )
 
 var (
--- a/demos/edgescroller/escroller.go
+++ b/demos/edgescroller/escroller.go
@@ -5,8 +5,8 @@ import (
 	"image/color"
 	"log"
 
-	"github.com/engoengine/engo"
 	"github.com/engoengine/ecs"
+	"github.com/engoengine/engo"
 )
 
 type Game struct{}
@@ -44,7 +44,6 @@ func generateBackground() *ecs.Entity {
 	bgTexture := engo.NewImageObject(img)
 	field := ecs.NewEntity([]string{"RenderSystem"})
 	fieldRender := engo.NewRenderComponent(engo.NewTexture(bgTexture), engo.Point{1, 1}, "Background1")
-	fieldRender.SetPriority(engo.Background)
 	fieldSpace := &engo.SpaceComponent{engo.Point{0, 0}, worldWidth, worldHeight}
 	field.AddComponent(fieldRender)
 	field.AddComponent(fieldSpace)
--- a/demos/zoom/zoom.go
+++ b/demos/zoom/zoom.go
@@ -5,8 +5,8 @@ import (
 	"image/color"
 	"log"
 
-	"github.com/engoengine/engo"
 	"github.com/engoengine/ecs"
+	"github.com/engoengine/engo"
 )
 
 type Game struct{}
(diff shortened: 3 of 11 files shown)


engo.go
```diff
@@ -20,7 +20,7 @@ var (
 	cam          *cameraSystem
 
 	scaleOnResize   = false
-	fpsLimit        = 120
+	fpsLimit        = 60
```
EtienneBruines: A lot of "normal" users will run at 60 fps.

demos/zoom/zoom.go
```diff
@@ -72,8 +71,8 @@ func (*Game) Type() string { return "Game" }
 func main() {
 	opts := engo.RunOptions{
 		Title:  "Zoom Demo",
-		Width:  400,
-		Height: 400,
+		Width:  900,
+		Height: 900,
```
EtienneBruines: This way you'll see the demo better.

camera.go
```diff
@@ -3,8 +3,8 @@ package engo
 import (
 	"sync"
 
-	"github.com/go-gl/mathgl/mgl32"
 	"github.com/engoengine/ecs"
+	"github.com/go-gl/mathgl/mgl32"
```
EtienneBruines: Apparently `goimports` was also called.

demos/edgescroller/escroller.go
```diff
@@ -44,7 +44,6 @@ func generateBackground() *ecs.Entity {
 	bgTexture := engo.NewImageObject(img)
 	field := ecs.NewEntity([]string{"RenderSystem"})
 	fieldRender := engo.NewRenderComponent(engo.NewTexture(bgTexture), engo.Point{1, 1}, "Background1")
-	fieldRender.SetPriority(engo.Background)
```
EtienneBruines: This was optional anyway.

paked:
These changes look good for me!

paked:
Thanks for the PR @EtienneBruines

PR #192, opened 2016-04-11
Add exit event, example, and ability to override default action

otraore:
Needs heavy review, modifications and comments are welcomed.


demos/animation/anim.go
```diff
@@ -96,6 +97,7 @@ func main() {
 		Title:  "Animation Demo",
 		Width:  1024,
 		Height: 640,
+		DefaultCloseAction,
```
EtienneBruines: This doesn't look valid.

otraore: Thank you fixing now

EtienneBruines: Also, having to set `true` to everything, when it's default, seems a bit redundant. Can you rename it such that the value `false` is the default? (without changing behavior? )

otraore: I was thinking the same thing, maybe instead have a function like 
```go
engo.SetDefaultCloseAction(false)
```

demos/hello/hello.go
```diff
@@ -3,10 +3,11 @@ package main
 import (
 	"image/color"
 	"log"
+	"fmt"
```
EtienneBruines: It adds `fmt` - either it is currently broken, or this breaks it.

demos/animation/anim.go
EtienneBruines: Or rename it to `DontExitOnClose` or something.

demos/hello/hello.go
otraore: Forgot to remove from testing, thanks for spotting

demos/scenes/scenes.go
```diff
@@ -30,7 +30,7 @@ func (game *IconScene) Setup(w *ecs.World) {
 	w.AddSystem(&ScaleSystem{})
 	w.AddSystem(&SceneSwitcherSystem{NextScene: "RockScene", WaitTime: time.Second * 3})
 
-	guy := ecs.NewEntity({"RenderSystem", "ScaleSystem")
+	guy := ecs.NewEntity("RenderSystem", "ScaleSystem")
```
EtienneBruines: This is indeed supposed to change; although it's already changed in the `master` branch.

otraore: It won't affect the merge request then, thanks again fro spotting!

engo_glfw.go
```diff
@@ -217,20 +224,44 @@ func RunPreparation(defaultScene Scene) {
...
 
 	ticker := time.NewTicker(time.Duration(int(time.Second) / fpsLimit))
-Outer:
-	for {
+	Outer: for {
 		select {
 		case <-ticker.C:
 			RunIteration()
-			if !headless && window.ShouldClose() {
+			if close {
```
EtienneBruines: I don't see where `close` is defined?

otraore: At the top

engo_glfw.go
```diff
@@ -28,11 +30,16 @@ var (
 	Hand      *glfw.Cursor
 	HResize   *glfw.Cursor
 	VResize   *glfw.Cursor
+	defaultCloseAction bool
+	game Scene
+	close bool
```
otraore: Close is defined here

engo_glfw.go
otraore: @EtienneBruines no, close is set when the game is going to be closed. ShouldClose and the signal call closeEvent().

engo_glfw.go
```diff
@@ -217,20 +224,44 @@ func RunPreparation(defaultScene Scene) {
 	WorldBounds.Max = Point{Width(), Height()}
 
 	SetScene(defaultScene, false)
+	game = defaultScene
+}
+
+func closeEvent() {
+	log.Println("Got close event")
+	game.Exit()
```
EtienneBruines: We can actually loop over all `Scene`s here, and call `Exit` on all of them?

otraore: Where are all the Scenes stored?

engo_glfw.go
EtienneBruines: Ah, I see it now :-).

engo_glfw.go
EtienneBruines: https://github.com/EngoEngine/engo/blob/master/scene.go#L10

And a `sceneWrapper` contains a `scene Scene`.

otraore: I will add that in the next commit

scene.go
```diff
@@ -25,6 +25,13 @@ type Scene interface {
 	// Hide is called when an other Scene becomes active
 	Hide()
 
+	// Exit is called when the user or the system requests to close the game
+	// This should be used to cleanup or prompt user if they're sure they want to close
+	// To prevent the default action (close/exit) make sure you set DefaultCloseAction in
+	// your game options to false and MAKE SURE to handle the exiting of the program by calling
+	//     engo.Exit()
+	Exit()
```
EtienneBruines: We should document it, whether or not we're calling it when the `Scene` is hidden / inactive

otraore: We should call it on all scenes regardless if it's hidden/inactive?

EtienneBruines: And they don't have to call `engo.Exit()`, they could also call` os.Exit()` if they wanted to.

EtienneBruines: I think we could, because it's "worse" than just hiding them. So even though it's hidden, they might want to do some cleanup / last-minute saving.

otraore: That's what I'm saying , we ARE calling it on all scenes, doesn't matter scene status

EtienneBruines: We [will be](https://github.com/EngoEngine/engo/pull/192/files#r59214946).

demos/exit/README.md
```diff
@@ -0,0 +1,10 @@
+# Exit Demo
```
paked: Yes! A demo!

demos/exit/exit.go
```diff
@@ -0,0 +1,97 @@
...
+func (game *GameWorld) Preload() {
+
+	// This could be done individually: engo.Files.Add("data/icon.png"), etc
+	// Second value (false) says whether to check recursively or not
+	engo.Files.AddFromDir("data", false)
+
+	log.Println("Preloaded")
+}
+
+func (game *GameWorld) Setup(w *ecs.World) {
```
paked: When writing a demo, I want to *only* have the functionality which that demo is about - in order to avoid confusion.

Basically, would you mind removing all of the rendering code?

otraore: @paked Sure!

engo_glfw.go
```diff
@@ -219,18 +225,44 @@ func RunPreparation(defaultScene Scene) {
 	SetScene(defaultScene, false)
 }
 
+func closeEvent() {
+	log.Println("Got close event")
```
paked: This log seems a bit redundant.

engo_glfw.go
```diff
@@ -219,18 +225,44 @@ func RunPreparation(defaultScene Scene) {
...
+	if defaultCloseAction {
+		log.Println("Closing")
+		Exit()
+	} else {
+		log.Println("Warning: default close action set to false, please make sure you manually handle this")
+	}
+}
+
 func runLoop(defaultScene Scene, headless bool) {
+	c := make(chan os.Signal, 1)
```
paked: Your indentation looks a bit wack here, would you mind running it through `goimports`/`gofmt`?

engo_glfw.go
otraore: It was a debug log, I will remove it

engo_glfw.go
```diff
@@ -219,18 +225,44 @@ func RunPreparation(defaultScene Scene) {
...
 	ticker := time.NewTicker(time.Duration(int(time.Second) / fpsLimit))
-Outer:
-	for {
+	Outer: for {
 		select {
 		case <-ticker.C:
 			RunIteration()
-			if !headless && window.ShouldClose() {
+			if close {
+				log.Println("Got close message from os signal")
```
paked: Again, lets get rid of these logs.

scene.go
```diff
@@ -25,6 +25,13 @@ type Scene interface {
 	// Hide is called when an other Scene becomes active
 	Hide()
 
+	// Exit is called when the user or the system requests to close the game
+	// This should be used to cleanup or prompt user if they're sure they want to close
+	// To prevent the default action (close/exit) make sure you call engo.OverrideDefaultAction() in
+	// your game options to false you should handle the exiting of the program by calling
+	//    engo.Exit()
+	Exit()
```
paked: :+1: For documentation!

engo_glfw.go
otraore: that's 4 spaces and my editor automatically does goimports/gofmt

paked: That looks like two tabs for me?

And there are a few other places where brackets are off (Look at your changes to the demos)

PR #202, opened 2016-04-17
Fix dependency documentation

### LINKED ISSUE #175: Document Windows-dependencies
I've managed to get it running by using [win-builds](http://win-builds.org/doku.php) and downloading every package, but @5k3105 was unable to do so with TDM. 

We should document that it works with the win-builds package (which include a lot, so I don't know which was the "magic fix").

paked:
Closes #175


EtienneBruines:
LGTM. 

Once brought to `master` we can merge this in.

paked:
@EtienneBruines Updated to master

PR #204, opened 2016-04-17
Add MasterVolume

### LINKED ISSUE #147: AudioSystem volume
It would be nice to have categories of audio fragments, and to be able to set the volume (loudness) for each of those categories, i.e. a category for ambiance, effects, voice, etc. These categories should be user-defined. 

We could also offer a master-volume.

paked:
Fix #147.

Up for review @EtienneBruines @otraore


audio.go
```diff
@@ -13,29 +14,41 @@ const (
...
 
+var MasterVolume float64 = 1
+
 // AudioComponent is a Component which is used by the AudioSystem
 type AudioComponent struct {
 	File       string
 	Repeat     bool
 	Background bool
 	player     *Player
+	RawVolume  float64
```
EtienneBruines: You may want to note that `RawVolume` shouldn't be changed? Or maybe we want to make sure they don' change it?

audio_player.go
```diff
@@ -329,11 +329,13 @@ func (p *Player) Volume() float64 {
 	if p == nil {
 		return 0
 	}
+
 	return float64(p.source.Gain())
 }
 
 // SetVolume sets the volume of the player. The range of the volume is [0, 1].
 func (p *Player) SetVolume(vol float64) {
+	fmt.Println("changing volume")
```
EtienneBruines: Debug logging, should be removed.

audio.go
```diff
@@ -78,6 +91,12 @@ func (as *AudioSystem) UpdateEntity(entity *ecs.Entity, dt float32) {
 		}
 	}
 
+	if MasterVolume != as.cachedVolume {
+		fmt.Println("Recalculating!", MasterVolume, as.cachedVolume)
```
EtienneBruines: Debug log; should be removed.

EtienneBruines:
Reviewed. Once you bring it up to master, we can theoretically merge it in. 

(I've enforced travis-build passing now)

EtienneBruines:
You have removed the `fmt` print statements, which is good. You however did not remove the `fmt` import statement, which is why the checks are failing.

paked:
:party:

PR #208, opened 2016-04-17
Added resource caching

### LINKED ISSUE #127: Preload and Scenes
Currently, one can override the behavior of "keep `World` intact". This is good. If chosen to create a new `World` instead, the `Setup` function is called - this is also good. 

However, the `Preload()` function is also called (another time). Is there a scenario in which we want this? (@paked) As files are usually just preloaded once.

EtienneBruines:
We can now call Preload as often as we'd like, because
resources that have been loaded already, will be re-used.

Fixes #127 

Ready for review and merge.

### DIFF
--- a/assets.go
+++ b/assets.go
@@ -92,34 +92,67 @@ func (l *Loader) Load(onFinish func()) {
 	for _, r := range l.resources {
 		switch r.kind {
 		case "png":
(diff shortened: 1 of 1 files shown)


otraore:
Looks good to me

Issue #211, opened 2016-04-18
Texture cleanup


EtienneBruines:
As noted in the comments at #135, we currently have no way to dispose of a `Texture`. A texture is being created on the GPU-side, but if we stop using our `RenderComponent` / `Texture`, that doesn't remove it from the GPU-side (causing memory leak). 

As Go doesn't have destructors, I believe we should just create some kind of cleanup function. Possible names:
* `Delete`
* `Remove`
* `Close` (yes, it's weird, but most people know they should `Close` whenever they can)

Thoughts everyone?

paked:
Lets use Close.

Issue #135, opened 2015-12-25
Font.Render cleanup


EtienneBruines:
Each time we use the `font.Render()` function to generate a `Texture`, the texture data is sent to the GPU. 

However, if the text we want to render, changes, the old `Texture` can be removed, if it's replaced by the new one. 

I am not sure what the best-practice is here; possibilities:
* `texture.Delete()` to be called whenever you're replacing the rendered text. 
* Create some kind of wrapper, called `Text`, which contains a reference to the correct font etc., and can be used to `Render` whenever the text-value changes, and removes the old `Texture` automatically. 

Thoughts @paked? We should do something, because this is leaking GPU memory.

paked:
`Text` is already a wrapper for `Font`. I don't think we need another one. If this behaviour should happen at the `Font` level, do so.

EtienneBruines:
We don't have something of type `Text`?

paked:
@EtienneBruines Looks like I implemented it, and then removed it at some point?

It definitely looks like something that should be implemented, as https://github.com/paked/engi/blob/master/demos/pong/pong.go#L266-L276. Does not look that great at all.

Ignore that previous message, implement it at the `Font` level... I'll create a wrapper once this Pull Request is through (I actually have time now, since holidays)

EtienneBruines:
Looking back at this issue; I believe we need some kind of cleanup-function for `Texture`s, not render-things specifically.

Issue #117, opened 2015-12-07
Golint engo


EtienneBruines:
Mostly a matter of "should have comment", but these should be fixed. 

```
engi_js.go:165:1:warning: exit is unused (deadcode)
audio_player.go:148::warning: cyclomatic complexity 17 of function (*Player).discoverHeader() is high (> 10) (gocyclo)
audio_player.go:187::warning: cyclomatic complexity 14 of function (*Player).prepare() is high (> 10) (gocyclo)
assets.go:94::warning: cyclomatic complexity 13 of function (*Loader).Load() is high (> 10) (gocyclo)
tmx_level.go:86::warning: cyclomatic complexity 13 of function createLevelFromTmx() is high (> 10) (gocyclo)
collision.go:72::warning: cyclomatic complexity 11 of function (*CollisionSystem).Update() is high (> 10) (gocyclo)
audio.go:60::warning: cyclomatic complexity 11 of function (*AudioSystem).Update() is high (> 10) (gocyclo)
animation.go:8:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
audio.go:6:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
engi.go:9:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
responder.go:8:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
scene.go:3:8:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
camera.go:11:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
collision.go:8:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
engi_glfw.go:23:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
render.go:8:2:error: could not import github.com/paked/engi/ecs (can't find import: github.com/paked/engi/ecs) (gotype)
responder.go:13:9:error: undeclared name: ecs (gotype)
scene.go:12:9:error: undeclared name: ecs (gotype)
animation.go:64:3:error: undeclared name: ecs (gotype)
animation.go:67:32:error: undeclared name: ecs (gotype)
[...]

PR #216, opened 2016-04-18
Rotation support

EtienneBruines:
Note: on account of preserving commits (and thus credit), let's not squash this one (even though we usually would). 

Let's review.


audio.go
```diff
@@ -53,7 +53,7 @@ func (as *AudioSystem) New(*ecs.World) {
 
 		// Hopefully not that much of an issue, when we receive it before the CameraSystem does
 		// TODO: but it is when the CameraMessage is not Incremental (i.e. the changes are big)
-		al.SetListenerPosition(al.Vector{cam.X() / Width(), cam.Y() / Height(), cam.Z() * as.HeightModifier})
+		al.SetListenerPosition(al.Vector{Cam.X() / Width(), Cam.Y() / Height(), Cam.Z() * as.HeightModifier})
```
EtienneBruines: Not sure again why this is being made public.

demos/falling/falling.go
```diff
@@ -156,8 +164,13 @@ func (*DeathSystem) Type() string { return "DeathSystem" }
 
 func (ds *DeathSystem) New(*ecs.World) {
 	// Subscribe to ScoreMessage
+<<<<<<< HEAD
+	engo.Mailbox.Listen("CollisionMessage", func(message engo.Message) {
+		collision, isCollision := message.(engo.CollisionMessage)
+=======
 	engo.Mailbox.Listen("ScoreMessage", func(message engo.Message) {
 		collision, isCollision := message.(engo.CollisionMessage)
+>>>>>>> origin
```
EtienneBruines: This ain't good

shaders.go
```diff
@@ -126,8 +179,17 @@ func (s *defaultShader) Draw(texture *webgl.Texture, buffer *webgl.Buffer, x, y,
...
 	}
 
-	// TODO: add rotation
+	// precomputed trigo on CPU to avoid GPU trigo for each vertex
+	angle := rotation * math.Pi / 180
+	c := math.Cos(angle)
+
+	Gl.Uniform1f(s.ufC, c)
+	Gl.Uniform1f(s.ufS, math.Sin(angle))
+	Gl.Uniform1f(s.ufOC, 1-c)
```
EtienneBruines: Not so happy about those `Uniform1f` - a lot of additional overload.

EtienneBruines:
# This breaks the `CameraSystem` and `CollisionSystem` so badly.

EtienneBruines:
I did some major reconstruction:
* Matrices everywhere
* Combined the two shaders to be the same, but with different options (`cameraEnabled` true/false). 

Stil breaking for the `CameraSystem` and `CollisionSystem` when rotating.

paked:
I'm going to have `do-not-merge` here until we decide whether or not we want to have the broken `{Collision,Camera}System` in `master`.

EtienneBruines:
# Current issues

* With rotation, we also have to rotate the 4 edges of the object, to see where they are, in order to detect whether or not we're hovering / colliding. 
* Allowing the choice between rotating around center, or rotating around `(0, 0)`. This further complicates the point above. 
* Some things just aren't square, even though we render them as such. Checking whether or not a pixel is transparent is easy without rotation, but becomes even more impossible / undoable when adding rotation.

EtienneBruines:
Not yet going as planned. 



The blue dot on the guy, is where `engo` thinks the mouse is hovering. The mouse is at the lower-right corner (where the screenshot ends).

EtienneBruines:
This PR is feature-complete. 

Ready for review.

(But don't merge until https://github.com/EngoEngine/engo/pull/247 is done!)

EtienneBruines:
Please review, anyone :smile:

camera.go
```diff
@@ -293,8 +319,36 @@ type MouseZoomer struct {
...
 		Mailbox.Dispatch(CameraMessage{Axis: ZAxis, Value: Mouse.ScrollY * c.ZoomSpeed, Incremental: true})
 	}
 }
+
+// MouseRotator is a System that allows for zooming when the scroll wheel is used
+type MouseRotator struct {
+	RotationSpeed float32
+
+	oldX    float32
+	pressed bool
```
EtienneBruines: We have to "remember" this, because currently we only receive events when the mouse changes (so one for pressing, and one for releasing, none for "holding").

shaders.go
```diff
@@ -106,193 +107,187 @@ void main (void) {
...
+	if s.cameraEnabled {
+		s.viewMatrix[1], s.viewMatrix[0] = math.Sincos(cam.angle * math.Pi / 180)
+		s.viewMatrix[3] = -s.viewMatrix[1]
+		s.viewMatrix[4] = s.viewMatrix[0]
+		s.viewMatrix[6] = -cam.x
+		s.viewMatrix[7] = -cam.y
+		s.viewMatrix[8] = cam.z
+	} else {
+		s.viewMatrix[6] = -1 / s.projectionMatrix[0]
+		s.viewMatrix[7] = 1 / s.projectionMatrix[4]
```
EtienneBruines: This notation (one line per index) is very verbose, but performs way better than reallocating a new `[]float32{}`.

shaders.go
```diff
@@ -106,193 +107,187 @@ void main (void) {
...
+	setBufferValue(buffer, 10, w, &changed)
+	setBufferValue(buffer, 11, h, &changed)
+	setBufferValue(buffer, 12, u2, &changed)
+	setBufferValue(buffer, 13, v2, &changed)
+	setBufferValue(buffer, 14, tint, &changed)
 
-	Gl.Uniform2f(s.ufPosition, x, y)
-	Gl.DrawElements(Gl.TRIANGLES, 6, Gl.UNSIGNED_SHORT, 0)
-}
+	//setValue(buffer, 15, 0, &changed)
```
EtienneBruines: Lines like these are commented because setting `0` is not needed (they're already 0) - but commented lines to provided the "bigger picture".

shaders.go
```diff
@@ -106,193 +107,187 @@ void main (void) {
...
+		buffer[index] = value
+		*changed = true
+	}
 }
 
 var (
-	DefaultShader = &defaultShader{}
-	HUDShader     = &hudShader{}
+	DefaultShader = &basicShader{cameraEnabled: true}
+	HUDShader     = &basicShader{cameraEnabled: false}
```
EtienneBruines: So this basically merges the two shaders into one shader, which results in nicer / maintainable code.

EtienneBruines:
Note to @paked and @otraore : This PR is completely ready to merge. 

Problems with MouseSystem / CollisionSystem have already been fixed. Also includes some code that helps with https://github.com/EngoEngine/engo/issues/237 and fixes #211 and fixes #135. 

Make sure (*if* you're going to squash), to give credit to @faide in that commit message, and to mention the two fixed issues and the one other issue that's partially fixed.

assets.go
```diff
@@ -175,6 +175,16 @@ func LoadShader(vertSrc, fragSrc string) *gl.Program {
...
 	Gl.CompileShader(vertShader)
+	var success int32
+	Gl.GetShaderiv(vertShader, Gl.COMPILE_STATUS, &success)
+	if success != int32(Gl.TRUE) {
+		var maxLength int32
+		Gl.GetShaderiv(vertShader, Gl.INFO_LOG_LENGTH, &maxLength)
+		log.Println("Error during vertex shader compilation")
+		errorLog := make([]byte, maxLength)
+		Gl.GetShaderInfoLog(vertShader, maxLength, &maxLength, (*uint8)(Gl.Ptr(errorLog)))
+		log.Print(string(errorLog))
```
otraore: Why not `log.Println` here?

camera.go
```diff
@@ -141,6 +154,10 @@ func (cam *cameraSystem) Z() float32 {
 	return cam.z
 }
 
+func (cam *cameraSystem) Angle() float32 {
```
otraore: I think we should try to document all exported functions when we can. Eventually there will be a PR addressing this but with new code we should try and document them as we go.

camera.go
```diff
@@ -293,8 +319,36 @@ type MouseZoomer struct {
 func (*MouseZoomer) Priority() int          { return 10 }
 func (*MouseZoomer) Remove(ecs.BasicEntity) {}
 
-func (c *MouseZoomer) Update(dt float32) {
+func (c *MouseZoomer) Update(float32) {
 	if Mouse.ScrollY != 0 {
+		// TODO: not sure if we have to use dt here, since we already use the "amount" of zooming
```
otraore: Just notifying you in case you need to come back and review this.

camera.go
```diff
@@ -293,8 +319,36 @@ type MouseZoomer struct {
...
 		Mailbox.Dispatch(CameraMessage{Axis: ZAxis, Value: Mouse.ScrollY * c.ZoomSpeed, Incremental: true})
 	}
 }
+
+// MouseRotator is a System that allows for zooming when the scroll wheel is used
```
otraore: Nice documentation on the struct

camera.go
```diff
@@ -293,8 +319,36 @@ type MouseZoomer struct {
...
+type MouseRotator struct {
+	RotationSpeed float32
```
otraore: RotationSpeed could use documentation (Not required)

camera.go
```diff
@@ -293,8 +319,36 @@ type MouseZoomer struct {
...
+
+	oldX    float32
+	pressed bool
+}
+
+func (*MouseRotator) Priority() int          { return 10 }
```
otraore: Why does this always return 10? Maybe set a constant and always return that?

collision.go
```diff
@@ -28,12 +29,46 @@ func (sc *SpaceComponent) Center(p Point) {
 }
 
 func (sc SpaceComponent) AABB() AABB {
```
otraore: I skipped mentioning documentation on a couple of other exported functions to not be extremely repetitive but this function really deserves one. I can't tell what it does from its contents.

collision.go
```diff
@@ -102,9 +139,13 @@ func (cs *CollisionSystem) Update(dt float32) {
 					e1.SpaceComponent.Position.Y += mtd.Y
 				}
 
+				collided = true
 				Mailbox.Dispatch(CollisionMessage{Entity: e1, To: e2})
 			}
 		}
+
+		// if no one collided reset to false
```
otraore: English: "if no one collided reset to false" -> "If nothing collided (re)set to false" makes just a bit more sense.

demos/mouse/mouse.go
```diff
@@ -27,7 +27,11 @@ func (*DefaultScene) Setup(w *ecs.World) {
 	w.AddSystem(&engo.MouseSystem{})
 	w.AddSystem(&engo.RenderSystem{})
 	w.AddSystem(&ControlSystem{})
+
+	// These are not required, but allow you to move / rotate and still see that it works
 	w.AddSystem(&engo.MouseZoomer{-0.125})
+	w.AddSystem(engo.NewKeyboardScroller(500, engo.W, engo.D, engo.S, engo.A))
```
otraore: The directions are inverted, W goes down, D goes up A goes right and S goes left for me. Also nothing special (rotation) happens when I hover/click (the hand shows up) What does engo.MouseZoomer do?

demos/mouserotation/rotation.go
```diff
@@ -0,0 +1,46 @@
...
+	rotationSpeed float32 = 10
+
+	worldWidth  int = 800
+	worldHeight int = 800
+)
+
+func (*DefaultScene) Preload() {}
+
+// Setup is called before the main loop is started
+func (*DefaultScene) Setup(w *ecs.World) {
```
otraore: Nothings happening for me I just see this screen:https://i.imgur.com/vHXgtf1.png  Am I missing anything?

demos/rotation/rotation.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+	if delete >= 0 {
+		r.entities = append(r.entities[:delete], r.entities[delete+1:]...)
+	}
+}
+
+func (r *RotationSystem) Update(dt float32) {
+	// speed in radians per second
+	var speed float32 = math.Pi
```
otraore: This seems useless. If `speed` is equal to `math.Pi` speedDegrees whcih is equal to `speed * 180 / math.Pi` doesn't make sense since it just ends up being `math.Pi/math.Pi` = 1 times 180 so speedDegrees will always be speedDegrees

assets.go
paked: Presumably OpenGL already has endlines on its logs?

camera.go
paked: We need to take another look at the way we handle priorities at some poitn.

assets.go
otraore: If that's the case then Ok.

demos/animation/anim.go
```diff
@@ -60,8 +60,12 @@ func (*DefaultScene) Type() string { return "GameWorld" }
...
 	entity := &Animation{BasicEntity: ecs.NewBasic()}
 
-	entity.SpaceComponent = engo.SpaceComponent{*point, 150, 150}
-	entity.RenderComponent = engo.NewRenderComponent(spriteSheet.Cell(0), engo.Point{3, 3})
+	entity.SpaceComponent = engo.SpaceComponent{
+		Position: *point,
+		Width:    150,
+		Height:   150,
+	}
+	entity.RenderComponent = engo.RenderComponent{Drawable: spriteSheet.Cell(0), Scale: engo.Point{3, 3}}
```
paked: Keeping consistent, could you split this struct intialisation over multiple lines (like `SpaceComponent`s)

demos/falling/falling.go
```diff
@@ -47,11 +47,11 @@ func (*DefaultScene) Setup(w *ecs.World) {
 	guy := Guy{BasicEntity: ecs.NewBasic()}
 
 	// Initialize the components, set scale to 4x
-	guy.RenderComponent = engo.NewRenderComponent(texture, engo.Point{4, 4})
+	guy.RenderComponent = engo.RenderComponent{Drawable: texture, Scale: engo.Point{4, 4}}
```
paked: Lets split this over multiple lines too... And for the rest of them

camera.go
EtienneBruines: Done.

camera.go
EtienneBruines: Done. Decided that we don't need `dt`, because we are already using "delta Y-coordinate".

camera.go
EtienneBruines: Done.

camera.go
EtienneBruines: @paked The only change that would make sense, is change it to `float32` (to allow decimals).

EtienneBruines: @otraore For now, I indeed use a constant.

collision.go
EtienneBruines: This one is indeed nice to document. Most of the rest will come whenever we decide to work on #117. 

Done.

collision.go
EtienneBruines: I decided to remove the comment altogether. Doesn't clarify that much, and it's only used internally.

demos/animation/anim.go
EtienneBruines: I will, as long as we set more than 1 variable. If we only set 0 or 1 variables, I'd rather keep it a one-liner.

demos/falling/falling.go
EtienneBruines: Done.

demos/mouse/mouse.go
EtienneBruines: `up, right, down, left`

So no, I shouldn't be passing `WASD`.

demos/mouserotation/rotation.go
EtienneBruines: Yes, you're not pressing the middle mouse button. It's already documented in the `README`.

demos/rotation/rotation.go
EtienneBruines: Yes and no. You can define speed either in radians or in degrees; this simply shows you can convert between the two - if needed.

EtienneBruines:
Processed all feedback. Feel free to merge whenever you guys agree :smile:

assets.go
mewmew: The `log` package automatically inserts newline characters.

From [log.Output](https://golang.org/pkg/log/#Output)
> A newline is appended if the last character of s is not already a newline.

EtienneBruines: Yeah, but if the string already contains a newline, `log.Println` will result in double newlines (since `log.Println` uses `fmt.Sprintln`,  which uses `fmt.doPrint(...)` which simply appends a newline. 

So if we were to use `log.Println`, it would result in a double newline in the logs. That's why we won't be using `Println` :smile: .

Thanks for doing the research though.

mewmew: > So if we were to use log.Println, it would result in a double newline in the logs. That's why we won't be using Println :smile: .

Confirmed in https://play.golang.org/p/bt3QhaNieM

demos/mouse/mouse.go
otraore: The directions are inverted on my computer

demos/mouserotation/rotation.go
otraore: I'm on a laptop, I don't have a middle mouse button, I'll try it out with my mouse later.

demos/rotation/rotation.go
otraore: I feel like this should be some form of a helper method. Something like math.Pitorad() of course it'd be a more welcoming function name.

assets.go
```diff
@@ -175,6 +175,16 @@ func LoadShader(vertSrc, fragSrc string) *gl.Program {
 	vertShader := Gl.CreateShader(Gl.VERTEX_SHADER)
 	Gl.ShaderSource(vertShader, vertSrc)
 	Gl.CompileShader(vertShader)
+	var success int32
+	Gl.GetShaderiv(vertShader, Gl.COMPILE_STATUS, &success)
+	if success != int32(Gl.TRUE) {
+		var maxLength int32
+		Gl.GetShaderiv(vertShader, Gl.INFO_LOG_LENGTH, &maxLength)
```
otraore: This breaks the gopherjs build some of the functions added here are not implemented on gopherjs.

EtienneBruines:
The squash/merge message should read something like this:

> Adds rotation support and texture cleanup
> 
> It also adds some features w.r.t. repeating textures, see issue #237. 
>
> Texture cleanup methods (`.Close()`) are added, which fixes #211 and fixes #135. 
>
> Shaders are reworked so they're easier to implement (also documented at #237). Buffers are now no longer handled by the `RenderComponent`. 
> 
> A lot of credit goes to @faide for doing the groundwork for the rotation support.

mewmew:
Lovely. Great work everyone!

paked:
Wooo!

PR #219, opened 2016-04-18
Handle Resources without extensions properly

### LINKED ISSUE #217: Resource without extension
Currently, whenever someone uses our asset manager to add a `Resource` which does not have a valid extension, it might panic. 

```go
func NewResource(url string) Resource {
	kind := path.Ext(url)
	//name := strings.TrimSuffix(path.Base(url), kind)
	name := path.Base(url)
	return Resource{name: name, url: url, kind: kind[1:]}
}
```

Specifically the `[1:]` part could panic if `len(kind)==0`. 

We should probably warn if someone adds a resource without extension (because we won't be able to load them anyway), but panic is a bit harsh.

paked:
Close #217

### DIFF
--- a/assets.go
+++ b/assets.go
@@ -1,6 +1,7 @@
 package engo
 
 import (
+	"fmt"
 	"image/color"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


assets.go
```diff
@@ -1,6 +1,7 @@
 package engo
 
 import (
+	"fmt"
```
EtienneBruines: Prefer `log` instead of `fmt` - we're already using `log` everywhere else.

PR #223, opened 2016-04-19
Removed std-lib math, fixes #222

### LINKED ISSUE #222: Migrate completely to luxengine/math
In relation to @newbrict's comment on #188.

We seem to have this idea of using this rewritten math package, however we have not actually fully migrated or committed to it yet. [A quick search](https://github.com/EngoEngine/engo/search?utf8=%E2%9C%93&q=%22math%22) shows that we are still using the stdlib's math package and converting to `float64` in a lot of places.

Let's fix this.

EtienneBruines:
The only place that currently uses std-lib math, is `engo_js.go`, but that's because we actually want `float64` math going on in there. 

So, basically all instances have been fixed now. 

Ready for review.


paked:
Looks like we broke the build :(

PR #226, opened 2016-04-19
Change engo.io/webgl to engo.io/gl

otraore:


EtienneBruines:
Travis likes it, then so do I.

PR #230, opened 2016-04-20
Made Exit, Hide and Show optional

### LINKED ISSUE #224: Exit optional
I've just been working on reworking `ecs`, and noticed that it's relatively easy to use interfaces to make certain functions optional. 

```go
type Exiter interface {
    Exit()
}

func whateverWeDoOnExit() {
    if exiter, ok := currentScene.(Exiter); ok {
        exiter()
    }
    // ...
}
```

This way, if they don't implement `Exit()`, it's OK (because no code is run). If they were to implement it, we would run it. 

As you made the current implementation, what's your thought on this, @otraore ?

### Show and Hide

Same goes for `Show()` and `Hide()` I guess, interfaces `Shower` and `Hider` would work.

EtienneBruines:
Fixes #224


paked:
LGTM

PR #240, opened 2016-04-21
Add information panic on font size >= 0

### LINKED ISSUE #238: Font Size panic
Whenever you omit the `Size` variable in the `engo.Font`, and call `Render`, it will panic (saying there's something wrong with slices). 

We should check if `Size == 0`, in our Render function. We may log / panic, but at least with a clear error message. 

I feel like panic is OK for this, because it's a programming-error, not a runtime-error per se.

paked:
Fix #238


EtienneBruines:
LGTM. Thank you for the fix.

PR #242, opened 2016-04-22
Fix RenderComponent constructor

paked:
Thank you @mewmew for pointing this out.


paked:
DNM Until demos are fixed.

paked:
ready to go!

Newbrict:
looks good to me, but why remove the label? what was the reason we needed it initially?

paked:
Response from Gitter.

> paked: @Newbrict I didn't remove the label, the label was removed a while back... Unfortunately it was left in the constructor.
> paked: It was originally used for associating resources with displaycomponents
> paked: If I remember correctly...

EtienneBruines:
LGTM. Thank you for fixing it. 

https://github.com/EngoEngine/engo/pull/216 would have also fixed it, but it's good to have it fixed before having to wait for such long-standing PR.

PR #244, opened 2016-04-23
Add default and looping animations

### LINKED ISSUE #45: Animation: default
The idea is that something has a user-defined default animation (i.e. standing still). 

Whenever someone wants to do something (i.e. animate a "grab", "jump", "kick", "nod", etc.), he calls a function and duration:
```go
something.Animate(ACTION_JUMP, time.Second * 0.5)
```
It animates as it normally would, but instead of repeating it, it would go back to animating the default behavior. 

Thoughts? @paked @matiwinnetou

paked:
Fixes #45.

### DIFF
--- a/animation.go
+++ b/animation.go
@@ -6,24 +6,26 @@ import (
 	"engo.io/ecs"
 )
 
-type AnimationAction struct {
+type Animation struct {
 	Name   string
 	Frames []int
+	Loop   bool
 }
 
 // Component that controls animation in rendering entities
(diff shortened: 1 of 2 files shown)


paked:
This also removes some of the "name" stutter in the function names.

otraore:
Looks good to me, LGTM.

EtienneBruines:
LGTM. I don't have the time personally to test if it actually works, but if it works, you may merge it in @paked :smile:

PR #247, opened 2016-04-24
[WIP] Gopherjs support

otraore:
This is still a work in progress and requires a lot of review and improvements.

Issue #245, opened 2016-04-23
Evaluate how a minimal engo API may look like


mewmew:
I'd like to evaluate which parts of `engo` make up the *core* of the engine. In particular, it would be interesting to find out whether we may arrive at a minimal API, from which common systems and components may be implemented as standalone packages, which would use composition, message-passing and prioritization of systems as provided by `ecs` and `engo`.

An initial evaluation has identified the six base components and six base systems, namely.

1. AnimationComponent
2. AudioComponent
3. CollisionComponent
4. MouseComponent
5. RenderComponent
6. SpaceComponent

and

1. AnimationSystem
2. AudioSystem
3. CameraSystem
4. CollisionSystem
5. MouseSystem
6. RenderSystem

I think it would be a show of strength if `engo` could be implemented as a set of decoupled packages using the principles behind ECS.

I am proposing that we evaluate the feasibility of extracting these systems and components out from `engo` and creating dedicated packages for them. It is too early to tell what effects this may have on the resulting API, and we may very well decide that a single monopackage is the way to go. Still, using this as a thought-experiment it would be interesting to see the dependencies between different parts of the `engo` engine brought into light by this evaluation.

An obvious place to start is to create a dedicated package for each component and system, but also one for the window handling, as `engo` may be used headless and a clean separation between the engine and its window handling would result in a much cleaner design.

What are your thoughts on this? Is it worth evaluating?

mewmew:
From commit mewpull/engo@131842414d833847eb2a89b2d992da4b372a3146.

> Initial separation of engo into small packages.
> 
> This is still very preliminary work. Follow-up commits
> will TODOs related to the split, and further decouple
> packages from one another.
>
> Updates EngoEngine/engo#245.

The *core* branch at https://github.com/mewpull/engo/tree/core is tracking preliminary work at evaluating what the minimal API of `engo` may look like, and whether it is possible to decouple components from one another.

mewmew:
A [modified version](https://github.com/mewspring/junk/tree/master/glyph) of [Dave Cheney's glyph command](https://godoc.org/github.com/davecheney/junk/glyph) was used to visualize the dependencies between the `engo` packages of the experimental *core* branch. The modified version of `glyph` excludes standard library packages from the graphs, to focus on the interesting parts, namely the intricate dependencies between different parts of `engo`.

Including some preliminary results below.

* Dependencies of the now minimal `engo` package and the recursive dependencies of its imports.

[...]

PR #251, opened 2016-04-25
Add generic asset loader to core

### LINKED ISSUE #246: Generic asset loader
It would be nice if users had the possibility to add their own filetype / extension / loader method to our asset loader. 

To be complete, we would then use that method ourselves. 

Thinking of a generic interface, let's brainstorm. 

```go
type FileLoader interface {
    // Extensions returns the file extensions for which this FileLoader should be used, excluding the '.' in front of it
    Extensions() []string

    // Load is called with the url (location) of the file, and is called whenever the file must be loaded
    Load(url string) error

    // Unload is called with the url (location) of the file, and is called whenever the resources of the file should be freed from the memory
    Unload(url string) error
}
```

There'd also have to be a generic method of retrieving the actual resource? Or not? Kinda like our `engo.Files.Image("url")` returns an actual (loaded) image. 

Thoughts? Improvements? Changes? Always welcome.

mewmew:
Fixes #246.


mewmew:
Don't merge this yet. There is still an ongoing discussion in #246 on how the API should look like. The good thing is, there will be no merge conflict as the `core` package doesn't exist yet, and it may happily live next to the `engo` package for quite some time as it evolves, and more and more features are decoupled from `engo` and implemented in `core`.

(We may still discuss what to name the `core` package. For now, it's a good placeholder name, until we find a better one)

otraore:
@mewmew +1 for documentation.

mewmew:
> @mewmew +1 for documentation.

@otraore thanks! I always try to tell myself that if it is difficult to document how an API is working, the API is probably too complex or wrong somehow.

paked:
@mewmew I think it might be beneficial to separate the move-to-core and generic-asset-loader parts of this pull request. [As stated at the bottom of #245](https://github.com/EngoEngine/engo/issues/245#issuecomment-213953471) I'd like to avoid splitting `core` and `engo` until #247 is merged.

core/files.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Files manages global resource handling of registered file formats for game
+// assets.
+var Files = &Formats{formats: make(map[string]FileLoader)}
+
+// FileLoader implements support for loading and releasing file resources.
+type FileLoader interface {
+	// Load loads the given resource into memory.
+	Load(url string) error
```
paked: Could we use an `io.Reader` here, and pass the URL in as a unique ID?

core/files.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Register registers a resource loader for the given file format.
+func (formats *Formats) Register(ext string, loader FileLoader) {
+	formats.formats[ext] = loader
+}
+
+// Load loads the given resource into memory.
+func (formats *Formats) Load(url string) error {
+	ext := filepath.Ext(url)
```
paked: I'd read in the contents of the `io.Reader` here, obviously this would be different on web and native (don't have to worry about that right now)

mewmew:
> edit: Read your comments more thoroughly above, still think these are two separate issues and should be treated as such.

Sure, no worries. I can move `files.go` to the `engo` package, and we may later decide where everything goes when the separation of `engo` and `core` happens.

core/files.go
EtienneBruines: That's probably the best reason for doing so. Such that *no-one* has to worry about stuff like that. We fix it once, and pass an `io.Reader` for them to use.

otraore:
Hey @mewmew I don't want this PR to go stale, so if you can please bring it up to master that'd be great!

mewmew:
> Hey @mewmew I don't want this PR to go stale, so if you can please bring it up to master that'd be great!

Updated to master. Still need to figure out where this goes. It currently lives in the core directory, as I've not removed the old assets loader yet. The generic assets loader should probably replace the old assets loader, in combination with the split to *core* packages, as both of these changes will require a lot of updates to all users of `engo`.

otraore:
Closing on request of @mewmew 

>I would expect that we simply close the #251 PR in favour of the core_api PR. 

from gitter

PR #254, opened 2016-04-26
check for lastBuffer for redraw

Newbrict:
this fixes reusing a texture like a tilemap atlas and changing offsets

### DIFF
--- a/shaders.go
+++ b/shaders.go
@@ -23,6 +23,7 @@ type defaultShader struct {
 	projY float32
 
 	lastTexture *gl.Texture
+	lastBuffer  *gl.Buffer
 
 	inPosition   int
 	inTexCoords  int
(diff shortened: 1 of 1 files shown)


shaders.go
```diff
@@ -23,6 +23,7 @@ type defaultShader struct {
 	projY float32
 
 	lastTexture *gl.Texture
+	lastBuffer *gl.Buffer
```
EtienneBruines: `gofmt`

shaders.go
```diff
@@ -127,7 +128,7 @@ func (s *defaultShader) Pre() {
 }
 
 func (s *defaultShader) Draw(texture *gl.Texture, buffer *gl.Buffer, x, y, rotation float32) {
-	if s.lastTexture != texture {
+	if s.lastTexture != texture || s.lastBuffer != buffer{
```
EtienneBruines: This goes into a separate `if`statement. 

```go
if s.lastBuffer != buffer {
    Gl.BindBuffer(Gl.ARRAY_BUFFER, buffer)
}
``` 

And then that line can be removed this `if` statement.

Newbrict:
don't merge

EtienneBruines:
LGTM. Thank you for your awesome work :smile: 

Will merge when travis is OK with it.

Issue #249, opened 2016-04-25
Add gopherjs to travis.yml


EtienneBruines:
It'd be nice if we tested the build integrity of `engo` w.r.t. gopherjs as well. 

Should be done after #247, though. 

Something I found on the web:
```
language: go
go:
  - 1.5
addons:
  apt:
    packages:
      - nodejs
install:
  - go get -u github.com/gopherjs/gopherjs
script:
  - diff -u <(echo -n) <(gofmt -d ./)
  - gopherjs test
```

I can imagine we also have to use that `script` tag to define it. Not sure if we have to write some weird script to find all `main` files, though. (Travis now does that for us, using just Go)

EtienneBruines:
Done in https://github.com/EngoEngine/engo/commit/e5a88d21b82a2fcab7048cf941b7bad02850c498

PR #255, opened 2016-04-27
Test demos in travis checks

otraore:
This will truly ensure that demos can be built when testing (this will be really useful for PRs).

https://www.reddit.com/r/golang/comments/3393hg/go_build_doesnt_build_binaries/

> When the command line specifies a single main package, build writes the resulting executable to output. Otherwise build compiles the packages but discards the results, serving only as a check that the packages can be built.
>


EtienneBruines:
You do know they are already being built by travis (by default) ? 

But I guess it helps with #249

mewmew:
Potentially, we could also make sure that the tutorial ([EngoEngine/TrafficManager](https://github.com/EngoEngine/TrafficManager)) is building, even if it is located in a separate repository.

.travis.yml
```diff
@@ -8,3 +8,6 @@ before_install:
 go_import_path: engo.io/engo
 go:
 - 1.6.2
+script:
+    - go test -v ./..
```
mewmew: Should be `go test -v ./...` (missing one dot `.`).

paked:
@mewmew That is a good idea. Lets get TrafficManager in on the travis-ci fun!

otraore:
"You do know they are already being built by travis (by default) ?" - @EtienneBruines 

To the best of my understanding, no they are not. Only tests are being ran as far as I know.

EtienneBruines:
@otraore I've seen tests fail (at this repo) because the demos weren't updated :P So I guess go test -v ./... does compile `main` packages.

otraore:
@EtienneBruines You're right but turns out it's go get. Anyones free to add in the TrafficManager to the build in the meantime, if not I'll do it tomorrow or something.

EtienneBruines:
Done in https://github.com/EngoEngine/engo/commit/e5a88d21b82a2fcab7048cf941b7bad02850c498.

PR #263, opened 2016-05-01
Renamed Width() and Height() to fix #212

### LINKED ISSUE #212: Naming of Width/Height WindowWidth/WindowHeight
The naming is currently horrible. 

We have two variants, both with width/height:
`WindowWidth` and `Width`. 

Currently, `WindowWidth` is a 1:1 match with the width in pixels of your game window. The `Width()` (which is the private variable `gameWidth`), is a bit more unclear. It's a 1:1 match with the width of your game window, **iff** `ScaleOnResize` is set to false. In other words, the `gameWidth` does not change if the window size changes. 

We should come up with better names, and document it better.

otraore:
Fixes https://github.com/EngoEngine/engo/issues/212#


EtienneBruines:
For future reference, I gave the `LGTM` on this one.

PR #266, opened 2016-05-02
Added initial legacy drawing support

EtienneBruines:
All done and ready for review (scroll down to latest `current state` comment).


EtienneBruines:
If anyone can comment as to which color the triangle is shown on the Triangle demo, that'd be awesome :smile: 

I currently see this

EtienneBruines:
Current state: Triangle is working.

EtienneBruines:
Curent status (in case I forget when I wake up tomorrow):

* Triangles can be drawn
* Rectangles can be drawn
* Circles can be drawn
* Circles rely on the window (Canvas!) height. 
* Circles have code that manually maps "ingame coordinates" to "screen coordinates", ugh...
* Nothing works with the `CameraSystem` yet. 
* I haven't tested rotation, but I fear that at least the circle will fail.

EtienneBruines:
Current status:

* Triangles, Rectangles, Circles and `ComplexTriangles` can be drawn.
* Color, size and rotation can be defined using `RenderComponent` and `SpaceComponent`. 
* They all work nicely with the `CameraSystem`, including rotation and zoom. 
* Borders are not yet operational. 

Is there anything else we'd like to see, order than the borders? @paked @otraore ?

otraore:
Nope this is perfect @EtienneBruines, If you can fix the travis builds, that should be it.

EtienneBruines:
Current status:

* Triangles, Rectangles, Circles and ComplexTriangles can be drawn. ComplexTriangles is a collection of points that make up triangles, which can be used to draw any object you want (as long as you can define it as a set of triangles).
* Color, size and rotation can be defined using RenderComponent and SpaceComponent.
* They all work nicely with the CameraSystem, including rotation and zoom.
* Borders are implemented
* Building is fixed (mainly because of @otraore)

Known bugs:
* `Circle` does not work with MSAA, and its border doesn't either. These have a worse quality than the others. 
* `ComplexTriangle` has a [weird-looking border](https://www.huperwebs.nl/upload/screenshot-2016-05-04_20:40:44.jpg) - no easy workaround AFAIK. 

My thought is to leave those bugs as-is, and create issues for them (I have no idea how to fix them, and wouldn't want this PR to be here forever). 

The coding part is done, so it's up for review.

engo_glfw.go
```diff
@@ -195,7 +197,7 @@ func SetBackground(c color.Color) {
 	if !headless {
 		r, g, b, a := c.RGBA()
 
-		Gl.ClearColor(float32(r), float32(g), float32(b), float32(a))
+		Gl.ClearColor(float32(r)/0xffff, float32(g)/0xffff, float32(b)/0xffff, float32(a)/0xffff)
```
EtienneBruines: This fixes the `engo.SetBackground` on the desktop. It currently only works with 100% black or 100% white. This allows for any color.

engo_js.go
```diff
@@ -36,7 +36,7 @@ func init() {
 //var canvas *js.Object
 var document = dom.GetWindow().Document().(dom.HTMLDocument)
 
-func CreateWindow(title string, width, height int, fullscreen bool) {
+func CreateWindow(title string, width, height int, fullscreen bool, msaa int) {
```
EtienneBruines: Not used, but still added to have the same signature as the desktop variant.

render.go
```diff
@@ -108,6 +108,8 @@ func (rs *RenderSystem) New(w *ecs.World) {
 		initShaders()
 	}
 
+	Gl.Enable(Gl.MULTISAMPLE)
```
EtienneBruines: No harm in doing this, even when `MSAA` is set to `1`.

shaders.go
```diff
@@ -122,7 +119,12 @@ void main (void) {
 }
 
 func (s *basicShader) Pre() {
+	// Enable shader and buffer, enable attributes in shader
 	Gl.UseProgram(s.program)
+	Gl.BindBuffer(Gl.ELEMENT_ARRAY_BUFFER, s.indexVBO)
+	Gl.EnableVertexAttribArray(s.inPosition)
+	Gl.EnableVertexAttribArray(s.inTexCoords)
+	Gl.EnableVertexAttribArray(s.inColor)
```
EtienneBruines: Best-practice is to `Enable` and `Disable` those whenever we switch shaders. If we don't, other shaders won't run without throwing errors.

shaders.go
```diff
@@ -209,6 +210,15 @@ func (s *basicShader) Draw(ren *RenderComponent, space *SpaceComponent) {
...
 	s.lastBuffer = nil
+
+	// Cleanup
+	Gl.DisableVertexAttribArray(s.inPosition)
+	Gl.DisableVertexAttribArray(s.inTexCoords)
+	Gl.DisableVertexAttribArray(s.inColor)
+
+	Gl.BindTexture(Gl.TEXTURE_2D, nil)
+	Gl.BindBuffer(Gl.ARRAY_BUFFER, nil)
+	Gl.BindBuffer(Gl.ELEMENT_ARRAY_BUFFER, nil)
```
EtienneBruines: These three lines unbind the buffers, so it won't be used by other shaders.

shaders.go
```diff
@@ -232,6 +242,10 @@ func (s *basicShader) generateBufferContent(ren *RenderComponent, buffer []float
 	h := ren.Drawable.Height()
 
 	colorR, colorG, colorB, colorA := ren.Color.RGBA()
+	colorR >>= 8
+	colorG >>= 8
+	colorB >>= 8
+	colorA >>= 8
```
EtienneBruines: Because `ren.Color.RGBA()` returns in the range of `0` to `0xffff` (instead of `0xff`), they need to be made smaller in order to work. This is being done in other parts of the code as well. 

Note this will result in a range from `0` to `255` (`0xff`). The `SetBackground` expected range from `0` to `1`, which is why we divided by `0xffff` in that scenario.

shaders.go
```diff
@@ -278,16 +292,489 @@ func setBufferValue(buffer []float32, index int, value float32, changed *bool) {
...
+varying vec4 var_BorderColor;
+
+void main (void) {
+  gl_FragColor = var_Color;
+
+  float halfBorder = var_BorderWidth / 2;
+
+  if (var_Radius.x > 0.0 && var_Radius.y > 0.0)
+  {
+    if (pow(gl_FragCoord.x - var_Center.x, 2.0) / pow(var_Radius.x - halfBorder, 2.0) + pow(gl_FragCoord.y - var_Center.y, 2.0) / pow(var_Radius.y - halfBorder, 2.0) > 1.0)
```
EtienneBruines: This line basically says: if this single pixel we're looking at, is outside of the circle (by looking at the range from the center of the circle), then ...

shaders.go
```diff
@@ -278,16 +292,489 @@ func setBufferValue(buffer []float32, index int, value float32, changed *bool) {
...
+    {
+      if (pow(gl_FragCoord.x - var_Center.x, 2.0) / pow(var_Radius.x + halfBorder, 2.0) + pow(gl_FragCoord.y - var_Center.y, 2.0) / pow(var_Radius.y + halfBorder, 2.0) > 1.0)
```
EtienneBruines: This line basically says: if the single pixel we're looking at, is outside of the circle (taking account the border as well), then ...

shaders.go
```diff
@@ -278,16 +292,489 @@ func setBufferValue(buffer []float32, index int, value float32, changed *bool) {
...
+	  {
+	    gl_FragColor.w = 0.0;
```
EtienneBruines: Make it invisible :smile:

shaders.go
```diff
@@ -278,16 +292,489 @@ func setBufferValue(buffer []float32, index int, value float32, changed *bool) {
...
+	  } else {
+	    gl_FragColor = var_BorderColor;
```
EtienneBruines: If not, show the border.

otraore:
LGTM, merging it in.

PR #271, opened 2016-05-05
fix some level stuff, add some math stuff

Newbrict:
@paked


level.go
```diff
@@ -1,18 +1,44 @@
 package engo
 
+import (
+	"engo.io/gl"
+)
+
 type Level struct {
-	Width      int
-	Height     int
+	width      int
```
paked: It'd be good to have a function to get width/height of the map in pixels?

level.go
```diff
@@ -1,18 +1,44 @@
...
+
+func (t *tile) Width() float32 {
+	return float32(t.width)
+}
+
+func (t *tile) Texture() *gl.Texture {
+	return t.Image.Texture()
+}
+
+func (t *tile) Close() {
```
paked: Don't we need to dump the region?

math.go
```diff
@@ -8,11 +8,31 @@ type Point struct {
 	X, Y float32
 }
 
+type Line struct {
```
paked: We're trying to make sure that all code that gets put into master now is documented. Could you do that here?

math.go
```diff
@@ -127,25 +142,111 @@ func (l *Line) PointDistanceSquared(point Point) float32 {
...
+	}
+
+	return Point{-1, -1}
+}
+
+// Returns the trace fraction of tracer through boundary
+// 1 means no intersection
+// 0 means tracer's origin lies on the boundary line
+func LineTraceFraction(tracer, boundary *Line) float32 {
+
```
paked: Don't need this newline

math.go
```diff
@@ -127,25 +142,111 @@ func (l *Line) PointDistanceSquared(point Point) float32 {
 		(y0-(y1+t*(y2-y1)))*(y0-(y1+t*(y2-y1)))
 }
 
-// Returns the point where the two lines intersect
-func (l *Line) LineIntersection(l2 Line) Point {
-	x1 := l.P1.X
-	x2 := l.P2.X
-	x3 := l2.P1.X
-	x4 := l2.P2.X
+// Left Hand Normal
```
paked: Can we please use godoc style comments: https://blog.golang.org/godoc-documenting-go-code

tmx_level.go
```diff
@@ -186,7 +186,7 @@ func pointStringToLines(str string, xOff, yOff float64) []Line {
 		floatPts[i][1], _ = strconv.ParseFloat(pt[1], 64)
 	}
 
-	lines := make([]Line, len(floatPts)-1)
+	lines := make([]*Line, len(floatPts)-1)
```
hydroflame: Memory discontinuation, don't do that

level.go
EtienneBruines: We should define the differences between them, and when to use which.

level.go
Newbrict: Yeah I am going to redo a bunch of the level stuff, the code changes to level and tile right now are just so that it's compliant with the current ECS implementation

level.go
Newbrict: You don't need to, I think in the end that's how it should be though. previous comment applies here too

level.go
EtienneBruines: > It'd be good to have a function to get width/height of the map in pixels?

As long as we don't ever call it "pixels" - because they're only pixels as long as the zoom-level is at 100% :smile: (and there's no scaling/resizing).

EtienneBruines:
LGTM. However: I haven't checked (1) the math, and (2) if it works. 

It would be great if someone created an issue where we added a `demo` which demonstrates this? 

And another issue for the math tests? 

(Anyone feel free to merge when it compiles; a second review is optional)

PR #273, opened 2016-05-05
Added fonts/Image.NRGBA for gopherjs

### LINKED ISSUE #260: Fonts on Web
Currently, gopherjs offers no font usage. 

Whenever you try to render a font, the game crashes. This is known, but should be fixed.

EtienneBruines:
Note you'll also have to update `engo.io/gl` for this. 

Fixes #260. 

No need for this to be around forever, it's feature complete and a small PR. Review and merge when ready :smile:


assets.go
```diff
@@ -316,3 +316,23 @@ func ImageToNRGBA(img image.Image, width, height int) *image.NRGBA {
...
+	return i.data
+}
+
+func (i *ImageObject) Width() int {
+	return i.data.Rect.Max.X
+}
+
+func (i *ImageObject) Height() int {
+	return i.data.Rect.Max.Y
+}
```
EtienneBruines: This specific `ImageObject` definition uses nothing more but standard library. We could use it in any platform we want (currently desktop *and* gopherjs).

engo_js.go
```diff
@@ -65,6 +65,8 @@ func CreateWindow(title string, width, height int, fullscreen bool) {
 	}
 	Gl.Viewport(0, 0, width, height)
 
+	Gl.GetExtension("OES_texture_float")
```
EtienneBruines: Required to directly load texture data.

engo_js.go
```diff
@@ -293,22 +295,22 @@ func loadFont(r Resource) (*truetype.Font, error) {
 	return freetype.ParseFont(ttfBytes)
 }
 
-type ImageObject struct {
+type HtmlImageObject struct {
```
EtienneBruines: Basically with the `texImage2D` method, you can either pass an HTML `` object (as we do with this type), or actual pixel data (which we do using `image.NRGBA` / `image.RGBA`).

otraore:
Great work @EtienneBruines and it looks ok except for one thing. Trying to enforce the use of comments on all exported functions, variables, types etc. If you can do that, that'd be sick.

EtienneBruines:
Done, so you may merge if you think it's ready :smile:

otraore:
LGTM Web is not officially supported! 🎉

Issue #261, opened 2016-04-30
Documentation of public API


mewmew:
Before the v1.0 release, the documentation of the public API of `engo` (and `core`) could use some love <3

This issue mirrors the one tracking `ecs` documentation at https://github.com/EngoEngine/ecs/issues/2

In particular, all exported functions, types, global variables, constants and methods should have a doc comment (of course using the conventions described in https://blog.golang.org/godoc-documenting-go-code).

Adding API examples would also be great (https://blog.golang.org/examples). If these were made runnable somehow from within the browser (using the GopherJS backend) that would be simply wonderful. I recon this may not be possible, but definitely worth investigating what the godoc sandbox allows and makes possible.

A more extensive package doc comment could be added to describe how the `engo` engine leverages the ECS paradigm to provide a set of core `System`s and `Component`s for implementing commonly used concepts in games.

This issue is related to the doc comment of the code. A separate issue could track the tutorial series and the concept explanations provided on engo.io.

otraore:
I put the on-hold label because this is one of those last few issues we deal with before 1.0. Upcoming PRs and changes will try to document anything they add in, but anything left at the end will be swooped in by this one.

otraore:
Update: Working on this

PR #283, opened 2016-05-09
Split into `engo` and `core`

### LINKED ISSUE #228: Revisit WorldBounds
Currently, we have `WorldBounds`, a global variable which allows you to set an `AABB` of where the game-world should be. These are being considered "holy" by the `CameraSystem` - you cannot move outside these bounds. 

The `WorldBounds` currently defaults to the window dimensions as defined in the `RunOpts`. 

Do we want to keep this behavior? It looks a bit weird in the demos, so maybe we should include setting the `WorldBounds` in the demo?

### LINKED ISSUE #235: engo.SetBackground
This really feels like an artifact of old engi

SetBackground should be part of the render system

### LINKED ISSUE #245: Evaluate how a minimal engo API may look like

### LINKED ISSUE #246: Generic asset loader

### LINKED ISSUE #253: Spritesheet has multiple textures
It would be nice if a spritesheet would just load one texture onto the GPU, and simply change the view. 

But I'm not sure if that's exactly what @Newbrict is trying to achieve at the moment?

### LINKED ISSUE #277: Update README with more accurate project status
I'd like to include in this:

- Status on mobile support
- Status on web support
- Status on Windows, OSX and Linux support
- Stability statement
- A preliminary release timeline

### LINKED ISSUE #280: Add tests for shaders
Currently, vertex shaders and texture shaders can fail. 

We should try to make some kind of `_test.go` file for those shaders, trying to compile them using some kind of headless OpenGL stuff. 

Up for grabs.

EtienneBruines:
According to #245. 

Also Fixes #245. Fixes #246. Fixes #253. Fixes #280. Fixes #277. Fixes #228. Fixes #235. 

Working towards #261 as well.


EtienneBruines:
It looks like #251 has to be in here as well. 

### Why?
`assets.go` defines different file formats, and thereby it defines what some types are that are only being used by `core`.

EtienneBruines:
Current state:

I moved most stuff to `core`. Currently left in `engo`:

* Window creation
* Asset loading / including definitions of Texture, NewTexture, ImageObject, etc. 
* Input handling
* Scenes
* Mailbox
* Math (`Point` and `AABB`)
* Cursor
* FPS calculator

# Current bugs:
* CameraSystem now isn't part of `engo` anymore, which means that we have to find a nice way for the `CameraSystem` to work with all other `System`s that **depend** on the `CameraSystem` (namely 2 shaders, and the `AudioSystem`)

assets.go
```diff
@@ -84,10 +82,6 @@ func (l *Loader) Json(name string) string {
 	return l.jsons[name]
 }
 
-func (l *Loader) Level(name string) *Level {
-	return l.levels[name]
-}
-
```
EtienneBruines: Removed but may later be added. It refers to `Level`, something available only in `core` (and this is `engo`, which cannot import `core`).

core/audio_unsupported.go
```diff
@@ -0,0 +1,36 @@
...
+	HeightModifier float32
+}
+
+func (as *AudioSystem) New(*ecs.World) {
+	log.Println("Warning: audio is not yet implemented on this platform")
+}
+
+func (as *AudioSystem) Add(*ecs.BasicEntity, *AudioComponent, *SpaceComponent) {}
+func (as *AudioSystem) Remove(basic ecs.BasicEntity)                           {}
+func (as *AudioSystem) Update(dt float32)                                      {}
```
EtienneBruines: I basically removed the `engo_js` and `engo_windows` methods. It basically says either "unsupported", or it's the default file which (in the future) will use `engo.io/audio`.

core/core.go
```diff
@@ -0,0 +1,2 @@
+// Package core contains the most-common System implementations according to the engo.io/ecs definition.
```
EtienneBruines: TODO: lookup syntax/references for "Package" documentation

EtienneBruines:
# Current state (in case I forget, or if anyone wants to contribute)

* Most stuff has been split
* assets.go contains a lot of commented code, that is usually code that needs to be ported / split. 
* core/imageLoader is done for the most part, some of the structs may have to be moved elsewhere (other files, maybe? )
* core/fontLoader still needs to implement the `Load` method, and I'm not sure about the `FontResource` definition yet. Not sure if it needs a `*gl.Texture` (which it does now), or just a `*truetype.Font`). 
* core/audioLoader still needs to be created
* Same goes for loading "tmx" files

We talked about doing this:
* A configurable "asset root directory", which is basically a user-definable string which you don't have to put in front of all your asset definitions. So if your file is stored at `assets/images/person.png`, then you may set `assets` as your "asset root directory", and reference that file by saying `images/person.png`. 
* This **will** default to a value of `assets`, because this is the directory-name `gomobile` **requires** in order to move those files into the `apk` file on *android*. 
* This should become part of `RunOpts` (defaulting to `assets` on an empty string) and an additional method to change it should also be provided (like with most `RunOpts` settings)

Afterwards, all demos have to be rewritten for new behavior. 

I will post again if I start working on this again. So feel free to contribute.

EtienneBruines:
Giving it another shot atm.

EtienneBruines:
# Current status

* All demos work
* It now fixes #253 as well (I got bored :laughing: )
* It now requires you to call `core.AudioSystemPreload()` before preloading any `.wav` files

Documentation leaves some things to be desired, but it is by no means worse than it is now. 

Please review changes so far, so we know what to improve. Feel free to commit the changes as well, if you'd like - but that's up to you. 

Currently **not** working on it, so feel free to.

core/core.go
mewmew: Not sure if this is what you were referring to?

From https://blog.golang.org/godoc-documenting-go-code

```go
// Package sort provides primitives for sorting slices and user-defined
// collections.
package sort
```

EtienneBruines: That is exactly what I'm referring to. Thank you @mewmew - I'm surprised I even did it right (by mistake)

mewmew:
> All demos work

That is absolutely wonderful, great work!!

> It now fixes #253 as well (I got bored :laughing: )

Haha, boredom must be the strongest force in the verse.

mewmew:
From rev 1ace0011895f718a5befbf06b777539509393f33

```
u@x1 ~/g/s/e/engo> go install
# engo.io/engo
./engo.go:195: undefined: warning
```

EtienneBruines:
Thank you @mewmew - forgot to push that one :smile:

mewmew:
> Thank you @mewmew - forgot to push that one :smile:

Thanks! It works great now. Had to update to the latest `engo.io/gl` package, and now all the demos work flawlessly. Really, you've done an amazing job!

assets.go
```diff
@@ -1,354 +1,138 @@
...
-	kind := path.Ext(url)
-	name := path.Base(url)
-
-	if len(kind) == 0 {
-		warning("Cannot load extensionless resource " + url)
-		return Resource{}
-	}
-
-	return Resource{name: name, url: url, kind: kind[1:]}
+type Resource interface {
```
mewmew: // A Resource represents a game resource, such as an image or a sound.

assets.go
```diff
@@ -1,354 +1,138 @@
...
+	URL() string
```
mewmew: Add from #251.

```go
// URL returns the uniform resource locator of the given resource.
```

assets.go
```diff
@@ -1,354 +1,138 @@
...
-
-			f, err := loadFont(r)
-			if err != nil {
-				log.Println("Error loading resource:", err)
-				continue // with other resources
-			}
-
-			l.fonts[r.name] = f
+// Load loads the given resource into memory.
+func (formats *Formats) Load(url string) error {
```
mewmew: Perhaps it doesn't matter, but `*Formats` does not implement the `FileLoader` interface anymore, as the signature for `Load` has been updated.

```go
Load(url string, data io.Reader) error
```

EtienneBruines: It shouldn't implement it, since it doesn't load any files? This is the "general hub" that calls all the other loaders? 

The reason we can't implement it, is because the `io.Reader` is "created" by `*Formats`.

mewmew: > It shouldn't implement it, since it doesn't load any files?

Sounds reasonable. Just wanted to make sure it was a decision we were aware about.

matiwinnetou:
I would suggest squashing commits to one via git rebase -i origin/master, it is kinda hard to review those commits retrospectively once they are in master.

mewmew:
> I would suggest squashing commits to one via git rebase -i origin/master, it is kinda hard to review those commits retrospectively once they are in master.

Sounds like a good idea. As far as I know, squashing commits is now part of the web interface of GitHub for pull requests, ref: https://github.com/blog/2141-squash-your-commits

EtienneBruines:
# Current status

* All tests pass (incl. shaders)
* All demos compile (desktop, gopherjs and mobile)
* The merge is complete (AFAIK)

* **It needs massive reviewing.**

I'm not working at it atm, so feel free to.

mewmew:
> It needs massive reviewing.

Will definitely try to take a more in-depth look at this once I get a chance. Have been quite busy at Uni lately.

README.md
```diff
@@ -34,6 +48,7 @@ us a DM or [create an issue](https://github.com/EngoEngine/engo/issues/new).
 Engo is currently undergoing a lot of optimizations and constantly gets new features. However, this sometimes means things break. In order to make transitioning easier for you, 
 we have a list of those changes, with the most recent being at the top. If you run into any problems, please contact us at [gitter](https://gitter.im/EngoEngine/engo). 
 
+* `engo` has been split in `engo` (which contains stuff about creating windows, starting the game, creating an OpenGL context, input handling, etc.) - and `core` (which contains a lot of common `System` implementations for common tasks (`RenderSystem`, `CameraSystem`, `AudioSystem`, etc.)
```
paked: We should probably explain this change at the top of the README so people understand the difference between `core` and `engo`.

> The top level `engo` package contains the functionality of creating windows, starting the game, working with an OpenGL context and handling input. It is deisgned to be used with Systems designed as per `engo.io/ecs` specifications. In `core` there are our ECS implementations of common game development Systems like a `RenderSystem`, `CameraSystem`.

Obviously that is a really rough draft, and you *could* probably copy and paste what you already have...

assets.go
paked: @mewmew @EtienneBruines Yes! This is what I was talking about in the issue a while back. Wonderful.

assets.go
```diff
@@ -1,354 +1,100 @@
...
-	return i.data.Rect.Max.X
-}
-
-// Height returns the maximum Y coordinate of the image
-func (i *ImageObject) Height() int {
-	return i.data.Rect.Max.Y
-}
+	log.Printf("no resource loader registered for file format %q", ext)
+	return nil, false
+}
```
paked: newline here please

clock.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/luxengine/math"
 )
 
+// A Clock is a measurement built in `engo` to measure the actual frames per seconds (framerate).
```
paked: Is this valid godoc?

core/audio.go
```diff
@@ -63,45 +82,59 @@ func (a *AudioSystem) Remove(basic ecs.BasicEntity) {
...
+	var cam *cameraSystem
+	for _, system := range w.Systems() {
+		switch sys := system.(type) {
+		case *cameraSystem:
+			cam = sys
+		}
+	}
+
+	if cam == nil {
+		log.Println("[ERROR] CameraSystem not found - have you added the `RenderSystem` before the `AudioSystem`?")
```
paked: This error could be quite confusing. Maybe it should be

> [ERROR] `CameraSystem` not found. You should add the `RenderSystem` before adding the `AudioSystem`.

clock.go
EtienneBruines: Yes. Words like "A", "An" and "The" can get added before the name.

paked: TIL...

core/audio.go
EtienneBruines: Hmm - but if they *did* add it before the `AudioSystem`, then it might confuse even more?

README.md
mewmew:
```diff
-In `core` there are our ECS implementations of common game development Systems like a `RenderSystem`, `CameraSystem`.
+The `core` package contains our ECS implementations of common game development Systems like a `RenderSystem`, `CameraSystem`.
```

EtienneBruines: @paked I think this change goes in the `Breaking Changes` section (where it is now), and in addition we explain what `engo` and what `core` is.

assets.go
EtienneBruines: As you wish.

clock.go
EtienneBruines: Yeah, I found out the day before that, thanks to @mewmew.

EtienneBruines:
# Current status

* `tmx_level.go` is currently being inactive, because it needs heavy work to work (and I have no idea what it's supposed to do, anyway). Unless someone really objects, I'd like to keep it that way for this PR and create an issue where we fix it. 
* Everything else works like a charm. It does as described in the first post in this PR. 

If anyone has reviewed it completely, please respond by a "thumbs up" or "LGTM" or something - that way this PR does not have to go stale.

mewmew:
> (I'd love to merge this in as soon as we can, so I won't have to either stop working on engo, or make this PR even large / harder to review)

I'll try my best to review it tonight or tomorrow :)

EtienneBruines:
> **@akiross**: (on Gitter)
> sure, I was questioning the name (`core`) though... Sounds like a `common`

I like the name `common` more than the name `core`. Thoughts everyone?

mewmew:
> I like the name common more than the name core. Thoughts everyone? 

Two characters longer, but works for me if everyone else agrees.

assets.go
```diff
@@ -1,354 +1,134 @@
...
-	}
-	for _, f := range files {
-		furl := url + "/" + f.Name()
-		if !f.IsDir() {
-			Files.Add(furl)
-		} else if recurse {
-			Files.AddFromDir(furl, recurse)
-		}
-	}
+	// Resource returns the given resource, and an error if it didn't succeed
```
mewmew: End sentence with period.

assets.go
```diff
@@ -1,354 +1,134 @@
...
-	return l.jsons[name]
-}
+// Formats manages resource handling of registered file formats.
+type Formats struct {
+	// formats maps from file extensions to resource loaders.
+	formats map[string]FileLoader
 
-func (l *Loader) Level(name string) *Level {
-	return l.levels[name]
+	// root is the directory which is prepended to every resource url internally
```
mewmew: End sentence with period.

assets.go
```diff
@@ -1,354 +1,134 @@
...
-	f, err := os.Open(l.sounds[name])
-	if err != nil {
-		return nil
-	}
-	return f
+// SetRoot can be used to change the default directory from `assets` to whatever you want.
+//
+// Whenever `root` does not start with the directory `assets`, you will not be able to support mobile (Android/iOS)
+// since they require you to put all resources within the `assets` directory. You can, however, use subfolders within
+// the `assets` folder, and set those as `root`.
```
mewmew: Add reference which states the requirement of mobile to place files in a directory called `assets`.

assets.go
```diff
@@ -1,354 +1,134 @@
...
-				log.Println("Error loading resource:", err)
-				continue // with other resources
-			}
-
-			l.fonts[r.name] = f
+// Load loads the given resource into memory.
+func (formats *Formats) Load(url string) error {
+	ext := filepath.Ext(url)
+	if loader, ok := Files.formats[ext]; ok {
+		readCloser, err := openFile(filepath.Join(formats.root, url))
```
mewmew: From https://github.com/golang/go/wiki/CodeReviewComments#variable-names
> Variable names in Go should be short rather than long. This is especially true for local variables with limited scope. Prefer c to lineCount.

Suggestion: rename `readCloser` to `f`.

assets.go
```diff
@@ -1,354 +1,134 @@
...
 }
 
-func ImageToNRGBA(img image.Image, width, height int) *image.NRGBA {
-	newm := image.NewNRGBA(image.Rect(0, 0, width, height))
-	draw.Draw(newm, newm.Bounds(), img, image.Point{0, 0}, draw.Src)
-
-	return newm
+// A ResourceLoaderNotFoundError is returned whenever the specified file format has no registered `FileLoader` to
+// return the requested resource.
+type ResourceLoaderNotFoundError struct {
```
mewmew: It seems to be conventional for error types to begin with `ErrFoo`.

Examples from https://golang.org/pkg/io/#pkg-variables
```go
var ErrClosedPipe = errors.New("io: read/write on closed pipe")
var ErrNoProgress = errors.New("multiple Read calls return no data or error")
var ErrShortBuffer = errors.New("short buffer")
var ErrShortWrite = errors.New("short write")
var ErrUnexpectedEOF = errors.New("unexpected EOF")
```

assets.go
EtienneBruines: With or without a whitespace after the period?

assets.go
mewmew: As per gitter discussion, remove exported error types unless they are required for a specific (real-world) use case. Use `fmt.Errorf` instead.

mewmew: Reference:

https://talks.golang.org/2014/readability.slide#14

> If client doesn't need to distinguish errors, e.g. ok with err != nil check only.
>
> ```go
fmt.Errorf("error in %s", val) or errors.New("error msg")
```
>
> If client wants to distinguish several errors by error code.
>
> ```go
var (
  ErrInternal   = errors.New("foo: inetrnal error")
  ErrBadRequest = errors.New("foo: bad request")
)
```
>
> If you want to put detailed information in error.
>
> ```go
type FooError struct { /* fields of error information */ }
func (e *FooError) Error() string { return /* error message */ }
&FooError{ /* set error data */ }
```

Comment (EtienneBruines) [inline assets.go]: That is a nice reference. The second option is the least desirable in the case of `VertexShaderCompilationError` (or whatever it is called), since we'd want to include the OpenGL error message if possible. 

The first one is indeed fine for the time being :smile:

File: assets.go [hunk 16]
```diff
@@ -1,354 +1,134 @@
...
+	Format string
+	URL    string
 }
 
-// ImageObject is a pure Go implementation of a `Drawable`
-type ImageObject struct {
-	data *image.NRGBA
+func (r ResourceLoaderNotFoundError) Error() string {
+	return fmt.Sprintf("no `FileLoader` registered for file format %s (within url: %s)", r.Format, r.URL)
```
mewmew: Personal preference, but I'd use `%q` rather than `%s` for the file extension and url, thus resulting in an output along the lines of:

```
no `FileLoader` registered for file format ".png" (within url: "assets/foo.png")
```

Note, for error messages (e.g. `r.Err`), I'd still use `%s`.

assets.go
```diff
@@ -1,354 +1,134 @@
...
+	// Err is the internal error of the `io` method used. Which method this is, is OS-dependant (`io` for desktop,
+	// `asset` for mobile, etc.)
+	Err error
 }
 
-// Height returns the maximum Y coordinate of the image
-func (i *ImageObject) Height() int {
-	return i.data.Rect.Max.Y
+func (r ResourceOpenError) Error() string {
+	return fmt.Sprintf("the assets manager was unable to open `Resource`: %s (%s)", r.URL, r.Err.Error())
```
mewmew: The `fmt` package is aware of the `error` interface, and much in the same way as you don't need to call `String` on `fmt.Stringer` values, you don't need to call `Error` on values implementing the `error` interface. It is good to know that values implementing both the `fmt.Stringer` and `error` interface, then `fmt` will invoke `Error` as the `error` interface takes precedence.

```diff
r.Err.Error()
r.Err
```

cursor.go
```diff
@@ -0,0 +1,15 @@
+package engo
+
+// Cursor is a reference to standard GLFW-cursors, to be used in conjunction with `SetCursor`
+type Cursor uint8
+
+const (
+	// CursorNone can be used to reset the cursor
+	CursorNone = iota
```
mewmew: Add explicit type to constant definitions. This will both make the documentation more pleasant as they are grouped with the relevant type, and improve type checking, as they would no longer be considered untyped constants.

```diff
-CursorNone = iota
+CursorNone Cursor = iota

File: engo_glfw.go [hunk 19]
```diff
@@ -419,41 +418,12 @@ func (i *ImageRGBA) Height() int {
...
-		return nil, err
-	}
-
-	return freetype.ParseFont(ttfBytes)
+// openFile is the desktop-specific way of opening a file
+func openFile(url string) (io.ReadCloser, error) {
+	return os.Open(url)
 }
 
+/*
```
mewmew: Remove out-commented code? Seems to be related to the old assets loader?

input_data.go
```diff
@@ -6,7 +6,9 @@ type Modifier int
 
 var (
 	Mouse mouse
+)
 
+const (
 	MOVE    = Action(0)
```
mewmew: Make constants more idiomatic Go by using camel case.

```diff
-MOVE
+Move
-PRESS
+Press
...
```

input_data.go
```diff
@@ -6,7 +6,9 @@ type Modifier int
 
 var (
```
mewmew: No need to use variable group for a single variable.

```diff
-var (
-   Mouse mouse
-)
+var Mouse mouse
```

input_data.go
mewmew: Although not strictly a change of this pull request, where we are here.

Define aliases using the constants they refer to rather than the value. This also make it possible to remove the comments without loss of detail.

```diff
-MouseButtonLeft   MouseButton = 0 // equivalent for MouseButton1
-MouseButtonRight  MouseButton = 1 // equivalent for MouseButton2
-MouseButtonMiddle MouseButton = 2 // equivalent for MouseButton3
+MouseButtonLeft   MouseButton = MouseButton1
+MouseButtonRight  MouseButton = MouseButton2
+MouseButtonMiddle MouseButton = MouseButton3
```

mewmew: Similarly as described above, use explicit types rather than inferred types when defining constants for keys.

```diff
- var (
- 	Dash         = Key(189)
- 	Apostrophe   = Key(222)
- 	Semicolon    = Key(186)
- 	Equals       = Key(187)
+ var (
+ 	Dash         Key = 189
+ 	Apostrophe   Key = 222
+ 	Semicolon    Key = 186
+ 	Equals       Key = 187
```

input_data.go
mewmew: The exported `Mouse` variable feels strange, as its type (`mouse`) is not exported, but the `mouse` type contains exported members (e.g. `mouse.X`, `mouse.Action`, etc). Is `engo.Mouse` intended to be exported? If so, its type should probably be exported as well, so it may be included in documentation.

math.go
```diff
@@ -4,6 +4,10 @@ import (
 	"github.com/luxengine/math"
 )
 
+type AABB struct {
```
mewmew: Add doc comment of AABB and its members.

math.go
```diff
@@ -4,6 +4,10 @@ import (
...
+	Min, Max Point
+}
+
 type Point struct {
```
mewmew: From `go doc image.Point`.

```go
// A Point is an X, Y coordinate pair. The axes increase right and down.
```

README.md
```diff
@@ -4,8 +4,22 @@
...
+([gopherjs](https://github.com/gopherjs/gopherjs)) is also available. 
+
+This table shows the current state of what's working and what's not. We are aiming to turn all **No** columns into **Yes** columns. 
+
+| Aspect/Target | Linux | Windows | OSX | WebGL | Android | iOS* |
+| ------------- | ----- | ------- | --- | ----- | ------- | ---  |
+| Compilation   | Yes   | Yes     | Yes | Yes   | Yes     | ?    |
+| AudioSystem   | Yes   | No      | Yes | No    | No      | ?    |
+| RenderSystem  | Yes   | Yes     | Yes | Yes   | Yes     | ?    |
+| XXXXXSystem   | Yes   | Yes     | Yes | Yes   | Yes     | ?    |
```
mewmew: XXXSystem. I'd love to see this implemented :)

scene.go
```diff
@@ -23,16 +23,22 @@ type Scene interface {
 	Type() string
 }
```
mewmew: Similarly, the `scenes` variable does not need a variable group as the group only contains a single variable.

```diff
- var (
- 	scenes = make(map[string]*sceneWrapper)
- )
+ var scenes = make(map[string]*sceneWrapper)
```

mewmew:
From running `u@x1 ~/g/s/e/engo> goimports -w .`

```patch
diff --git a/core/audio.go b/core/audio.go
index a124e0e..b9a0d9d 100644
--- a/core/audio.go
+++ b/core/audio.go
@@ -5,10 +5,11 @@ package core
 import (
 	"log"
 
+	"io"
+
 	"engo.io/ecs"
 	"engo.io/engo"
 	"golang.org/x/mobile/exp/audio/al"
-	"io"
 )
 
 const (
diff --git a/core/audio_unsupported.go b/core/audio_unsupported.go
index 9188948..a0269af 100644
--- a/core/audio_unsupported.go
+++ b/core/audio_unsupported.go
@@ -3,8 +3,9 @@
 package core
 
 import (
-	"engo.io/ecs"
 	"io"
+
+	"engo.io/ecs"
 )
 
 const (
diff --git a/core/mouse.go b/core/mouse.go
index 35446eb..ca5c822 100644
--- a/core/mouse.go
+++ b/core/mouse.go
@@ -85,7 +85,7 @@ type mouseEntity struct {
 type MouseSystem struct {
 	entities []mouseEntity
 	world    *ecs.World
-	camera *cameraSystem
+	camera   *cameraSystem
 
 	mouseX    float32
 	mouseY    float32
@@ -106,7 +106,7 @@ func (m *MouseSystem) New(w *ecs.World) {
 		}
 	}
 
-	if m.camera== nil {
+	if m.camera == nil {
 		log.Println("ERROR: CameraSystem not found - have you added the `RenderSystem` before the `MouseSystem`?")
 		return
 	}
diff --git a/core/render.go b/core/render.go
index d577f0f..027f6b5 100644
--- a/core/render.go
+++ b/core/render.go
@@ -189,9 +189,9 @@ func (rs *RenderSystem) Update(dt float32) {
 }
 
 func SetBackground(c color.Color) {
-	if !engo.Headless(){
+	if !engo.Headless() {
 		r, g, b, a := c.RGBA()
 
-		engo.Gl.ClearColor(float32(r) / 0xffff, float32(g) / 0xffff, float32(b) / 0xffff, float32(a) / 0xffff)
+		engo.Gl.ClearColor(float32(r)/0xffff, float32(g)/0xffff, float32(b)/0xffff, float32(a)/0xffff)
 	}
-}
\ No newline at end of file
+}
diff --git a/core/render_filetype.go b/core/render_filetype.go
index 4defc57..874b7af 100644
--- a/core/render_filetype.go
+++ b/core/render_filetype.go
@@ -6,9 +6,10 @@ import (
 	_ "image/png"
 	"io"
 
+	"fmt"
+
 	"engo.io/engo"
 	"engo.io/gl"
-	"fmt"
 )
 
 type TextureResource struct {
diff --git a/core/render_shaders.go b/core/render_shaders.go
index 367537d..f59fa8e 100644
--- a/core/render_shaders.go
+++ b/core/render_shaders.go
@@ -3,13 +3,14 @@ package core
 import (
 	"image/color"
 
+	"fmt"
+	"log"
+	"strings"
+
 	"engo.io/ecs"
 	"engo.io/engo"
 	"engo.io/gl"
-	"fmt"
 	"github.com/luxengine/math"
-	"log"
-	"strings"
 )
 
 const bufferSize = 10000
diff --git a/core/render_shapes.go b/core/render_shapes.go
index afdd7d6..0b903d2 100644
--- a/core/render_shapes.go
+++ b/core/render_shapes.go
@@ -1,9 +1,10 @@
 package core
 
 import (
+	"image/color"
+
 	"engo.io/engo"
 	"engo.io/gl"
-	"image/color"
 )
 
 type TriangleType uint8
diff --git a/core/tmx_level.go b/core/tmx_level.go
index 55eead2..8ebb72f 100644
--- a/core/tmx_level.go
+++ b/core/tmx_level.go
@@ -8,13 +8,14 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/xml"
-	"engo.io/engo"
 	"fmt"
 	"io/ioutil"
 	"path"
 	"sort"
 	"strconv"
 	"strings"
+
+	"engo.io/engo"
 )
 
 // Just used to create levelTileset->Image
diff --git a/demos/audio/audio.go b/demos/audio/audio.go
index 524ea6d..4c5947e 100644
--- a/demos/audio/audio.go
+++ b/demos/audio/audio.go
@@ -3,10 +3,11 @@ package main
 import (
 	"image/color"
 
+	"log"
+
 	"engo.io/ecs"
 	"engo.io/engo"
 	"engo.io/engo/core"
-	"log"
 )
 
 type DefaultScene struct{}
diff --git a/demos/hide/hide.go b/demos/hide/hide.go
index 6960a2a..83e1142 100644
--- a/demos/hide/hide.go
+++ b/demos/hide/hide.go
@@ -4,10 +4,11 @@ import (
 	"image/color"
 	"math/rand"
 
+	"log"
+
 	"engo.io/ecs"
 	"engo.io/engo"
 	"engo.io/engo/core"
-	"log"
 )
 
 type DefaultScene struct{}
diff --git a/engo.go b/engo.go
index a61079c..80af889 100644
--- a/engo.go
+++ b/engo.go
@@ -28,7 +28,7 @@ var (
 const (
 	// DefaultVerticalAxis is the name of the default vertical axis, as used internally in `engo` when `StandardInputs`
 	// is defined.
-	DefaultVerticalAxis   = "vertical"
+	DefaultVerticalAxis = "vertical"
 
 	// DefaultHorizontalAxis is the name of the default horizontal axis, as used internally in `engo` when `StandardInputs`
 	// is defined.
```

core/render.go
```diff
@@ -1,11 +1,12 @@
...
 
 import (
 	"fmt"
 	"image/color"
 	"sort"
 
 	"engo.io/ecs"
+	"engo.io/engo"
 	"engo.io/gl"
 )
```
mewmew: Use more idiomatic names for texture repeating constants.

```diff
 const (
- 	CLAMP_TO_EDGE TextureRepeating = iota
- 	CLAMP_TO_BORDER
- 	REPEAT
- 	MIRRORED_REPEAT
+ 	ClampToEdge TextureRepeating = iota
+ 	ClampToBorder
+ 	Repeat
+ 	MirroredRepeat
 )
```

core/render_filetype.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (r Texture) Close() {
+	if !engo.Headless() {
+		engo.Gl.DeleteTexture(r.id)
+	}
+}
+
+func init() {
+	engo.Files.Register(".jpg", &imageLoader{images: make(map[string]TextureResource)})
```
mewmew: Beautiful :)

mewmew:
Once all the comments have been addressed you definitely have a LGTM for this PR. Absolutely wonderful job @EtienneBruines!

We can always continue to improve and hopefully pull out even a few parts from `engo`, or unexport details not relevant to the users. For now, lets make sure to get this merged in soon, so we may all build upon and improve the post-split version of Engo!

Good night everyone :)

README.md
paked: Yep!

input_data.go
EtienneBruines: You make a valid point. 

It is in fact meant to be exported. I'm not sure what the new name of `mouse` would become. 

Perhaps move `engo.Mouse` to `engo.Input.Mouse`? Or something like that? (@paked? )

cursor.go
EtienneBruines: Can't believe I forgot that :open_mouth:

input_data.go
EtienneBruines: Chose to export `mouse` after renaming it to `MouseState`. (since it holds information about the current state of the mouse). 

If we ever want to move it to `engo.Input`, then that's for another PR.

EtienneBruines:
The squash commit message would be something like this:

```
Split engo into engo and common

By doing that, it fixes #245. 

A generic asset loader was implemented, which works for every platform, every file. Each platform now only has to implement a openFile method to support those files. This fixes #246. Thanks to @mewmew for the initial code. 

Some spritesheet optimization was done, to ensure a Spritesheet is just one Texture on the GPU-side of things. Fixes #253. 

Shader-compilation is now being tested (except for Windows, due to limitations of AppVeyor). Fixes #280. 

The README now has a bit more accurate information on the current state, and fixed some bugs and typos (thanks @otraore). Fixes #277. 

Moved WorldBounds to common.CameraBounds to make its usage clearer. Fixes #228. 

SetBackground was moved to common, and fixes #235 in doing so. 

A little bit more documentation was added, which helps with #261. 

Thanks to everyone for the discussion on this. As well as everyone who has helped review. This was a collaborative effort. 
```

mewmew:
:tada:

Issue #291, opened 2016-05-13
Move to luxengine/glm


otraore:
Our current math implementation is duplicating and not as accurate as it should be. No need to try and improve it when there's an implementation by @hydroflame and @luxengine. I'm proposing we make the move to it after the core_api PR is merged in,

EtienneBruines:
luxengine/glm: https://github.com/luxengine/glm

Reasons for this also include the fact that the `glm` package provided by @luxengine has extensive tests for floating-point precision correctness. Something that tends to go wrong often, when working with floats and math).

Newbrict:
this is dank

otraore:
Would we consider https://github.com/EngoEngine/engo/commit/8cf20599912cd81757e1e6b1312a5624ae3dd413 a fix for this?

EtienneBruines:
No. That was just to not make it worse (and to actually fix the bug).

Noofbiz:
Is this still open? From the looks of things we've moved to the EngoEngine fork of this a while back. Should this be closed?

otraore:
Yeah, this is outdated, closing.

PR #298, opened 2016-05-13
Update README.md

Newbrict:


otraore:
Please remove vector math because that's being changed to https://github.com/EngoEngine/engo/issues/291

PR #301, opened 2016-05-15
Fixed coveralls

otraore:


otraore:
Merging in since it doesn't break anything, can always be reverted.

PR #304, opened 2016-05-15
Move audio to an audio package and add support for web

otraore:
Work in progress. Also depends on engo.io/audio package which is also a work in progress. 
Status:
Desktop/Native sounds the same (nothing broke as far as I know)
Web sounds also work, but we're making a move to http://www.html5rocks.com/en/tutorials/webaudio/positional_audio/ in the next few days on the audio package


assets.go
```diff
@@ -48,6 +48,11 @@ func (formats *Formats) SetRoot(root string) {
 	formats.root = root
 }
 
+// GetRoot returns the currently set root directory for assets
+func (formats *Formats) GetRoot() string {
```
EtienneBruines: Idiomatic Go would name this `Root() string`.

common/audio_desktop.go
```diff
@@ -0,0 +1,123 @@
+//+build !windows,!netgo,!android
```
EtienneBruines: I guess in the future (as in: before merging), the `engo.io/audio` API would be such that we don't need an `audio_desktop.go` file here, but just one `audio.go` file?

demos/hello-world/hello.go
```diff
@@ -42,7 +42,7 @@ func (*DefaultScene) Setup(w *ecs.World) {
 		Scale:    engo.Point{8, 8},
 	}
 	guy.SpaceComponent = common.SpaceComponent{
-		Position: engo.Point{0, 0},
+		Position: engo.Point{1, 1},
```
EtienneBruines: Why?

otraore: Sorry this was me testing something when we had that vanvasWidth/Height issue.

common/audio_desktop.go
otraore: Hopefully, will be a little tricky though because of openal but I have an idea in mind.

Noofbiz:
Closed by #494

PR #309, opened 2016-05-17
Fix package name in README.md for hud demo

jzs:
Fix for README file that wasn't updated to reflect the new package changes


EtienneBruines:
Thank you for spotting this!

PR #311, opened 2016-05-17
Fixed #310

### LINKED ISSUE #310: MouseSystem: Hover with rotation
In this example, the green rectangle is where the cursor is located. It's green because `engo` thinks we're hovering the road (gray diagonal line). 

The road is created by taking a 200x20 rectangle, and rotating that at some angle.

EtienneBruines:
`SpaceComponent.AABB` is now no longer used (but still exists
for those that want to use it).

We now use and offer a `Within` method, checking whether or not
a given point is within the bounds of the `SpaceComponent`,
taking into account things as `Rotation` and floating precision.

It is feature-complete / done, so please review and merge. :smile:

### DIFF
--- a/common/collision.go
+++ b/common/collision.go
@@ -5,6 +5,7 @@ import (
 
 	"engo.io/ecs"
 	"engo.io/engo"
+	"github.com/luxengine/glm"
 	"github.com/luxengine/math"
 )
 
@@ -43,27 +44,27 @@ func (sc SpaceComponent) AABB() engo.AABB {
 	corners := sc.Corners()
 
 	var (
-		X_MIN float32 = -math.MaxFloat32
-		X_MAX float32 = math.MaxFloat32
-		Y_MIN float32 = -math.MaxFloat32
-		Y_MAX float32 = math.MaxFloat32
+		xMin float32 = -math.MaxFloat32
+		xMax float32 = math.MaxFloat32
+		yMin float32 = -math.MaxFloat32
+		yMax float32 = math.MaxFloat32
 	)
 
 	for i := 0; i < 4; i++ {
-		if corners[i].X < X_MIN {
-			X_MIN = corners[i].X
-		} else if corners[i].X > X_MAX {
-			X_MAX = corners[i].X
+		if corners[i].X < xMin {
+			xMin = corners[i].X
+		} else if corners[i].X > xMax {
+			xMax = corners[i].X
 		}
-		if corners[i].Y < Y_MIN {
-			Y_MIN = corners[i].Y
+		if corners[i].Y < yMin {
+			yMin = corners[i].Y
 		}
-		if corners[i].Y > Y_MAX {
-			Y_MAX = corners[i].Y
+		if corners[i].Y > yMax {
+			yMax = corners[i].Y
 		}
 	}
 
-	return engo.AABB{engo.Point{X_MIN, Y_MIN}, engo.Point{X_MAX, Y_MAX}}
+	return engo.AABB{engo.Point{xMin, yMin}, engo.Point{xMax, yMax}}
 }
 
 // Corners returns the location of the four corners of the rectangular plane defined by the `SpaceComponent`, taking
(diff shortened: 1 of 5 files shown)


common/collision_test.go
```diff
@@ -0,0 +1,58 @@
...
+		assert.True(t, exp1[i].Equal(act1[i]), fmt.Sprintf("corner %d did not match for rotation %f (got %v expected %v)", i, space1.Rotation, act1[i], exp1[i]))
+	}
+
+	space2 := SpaceComponent{Width: 1, Height: 1, Rotation: 90}
+	exp2 := [4]engo.Point{engo.Point{0, 0}, engo.Point{0, 1}, engo.Point{-1, 0}, engo.Point{-1, 1}}
+	act2 := space2.Corners()
+	for i := 0; i < 4; i++ {
+		assert.True(t, exp2[i].Equal(act2[i]), fmt.Sprintf("corner %d did not match for rotation %f (got %v expected %v)", i, space2.Rotation, act2[i], exp2[i]))
+	}
+}
```
EtienneBruines: We could add some more tests in the future, but for now this is OK.

common/mouse.go
```diff
@@ -199,10 +199,9 @@ func (m *MouseSystem) Update(dt float32) {
 		// If the Mouse component is a tracker we always update it
 		// Check if the X-value is within range
 		// and if the Y-value is within range
-		pos := e.SpaceComponent.AABB()
 
 		if e.MouseComponent.Track || e.MouseComponent.startedDragging ||
-			mx > pos.Min.X && mx < pos.Max.X && my > pos.Min.Y && my < pos.Max.Y {
+			e.SpaceComponent.Within(engo.Point{mx, my}) {
```
EtienneBruines: Looks nicer already :smile:

common/collision.go
```diff
@@ -37,33 +40,84 @@ func (sc SpaceComponent) AABB() engo.AABB {
...
+	return engo.AABB{engo.Point{X_MIN, Y_MIN}, engo.Point{X_MAX, Y_MAX}}
+}
+
+// Corners returns the location of the four corners of the rectangular plane defined by the `SpaceComponent`, taking
+// into account any possible rotation.
+func (sc SpaceComponent) Corners() (points [4]engo.Point) {
+	points[0].X = sc.Position.X
+	points[0].Y = sc.Position.Y
+
+	sin, cos := math.Sincos(sc.Rotation * math.Pi / 180)
```
EtienneBruines: You would think that `math.Sincos` can be skipped if `Rotation == 0`, and that is true.

However, reliably checking whether `Rotation == 0`, requires additional resources which make it not worth it.

otraore:
If you can try and fix the appveyor build I'll review it later on.

common/collision_test.go
otraore: Fails on windows (appveyor)

common/mouse.go
paked: Within sounds like it would be a point->AABB reaction. I think we should rename it.

common/collision.go
```diff
@@ -37,33 +40,84 @@ func (sc SpaceComponent) AABB() engo.AABB {
...
-	sin, cos := math.Sincos(sc.Rotation * math.Pi / 180)
-	xmin := sc.Position.X
-	xmax := sc.Position.X + sc.Width*cos - sc.Height*sin
-	ymin := sc.Position.Y
-	ymax := sc.Position.Y + sc.Height*cos + sc.Width*sin
+	corners := sc.Corners()
 
 	var (
-		X_MIN, X_MAX, Y_MIN, Y_MAX float32
+		X_MIN float32 = -math.MaxFloat32
```
paked: These are not idiomatic. Should be `XMin`, etc.

common/mouse.go
EtienneBruines: We could rename it, but it's definitely *not* a `point->AABB` reaction. `AABB` kind defines the "outer bounds" by defining  non-rotated rectangle. 

The green is what's defined by AABB, and that's *not* what's being used here.

EtienneBruines:
@otraore Fixed AppVeyor. Have fun reviewing :smile:

otraore:
LGTM. As for https://github.com/EngoEngine/engo/pull/311/files#r63638368 We'll merge it in for now and if @paked still shares the same view, we can always rename it.

PR #314, opened 2016-05-18
cameraSystem is now CameraSystem (exported)

### LINKED ISSUE #312: CameraSystem unexported, cannot follow an entity
As @TheBerryBeast has rightfully noted, because the `common.cameraSystem` is currently unexported, it is not possible to make it follow an entity. 

We should make it exported for this reason.

EtienneBruines:
Fixes #312


otraore:
LGTM

PR #315, opened 2016-05-19
Added support for tmx file loading

Newbrict:


common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
+package common
+
+import (
+	"bytes"
+	"fmt"
+        "engo.io/engo"
+	"io"
+)
+
+// TMXResource is a wrapper for a level that was created from a Tile Map XML
```
otraore: Better documentation please, this line sounds funny

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+type TMXResource struct {
+	Level *Level
+	url   string
+}
+
+func (r TMXResource) URL() string {
```
otraore: Document this

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+	return r.url
+}
+
+// tmxLoader is responsible for managing '.tmx' files within 'engo.Files'
```
otraore: +1 for documentation

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+type tmxLoader struct {
+	tmxs map[string]TMXResource
+}
+
+// Load will read the tmx file into a string to create a level using createLevelFromTmx.  Any image files required for creating the level will be loaded in as needed.
```
otraore: Split this into multiple lines and I don't think you need to mention createLevelFromTmx

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+	lvl, err := createLevelFromTmx(buf.String())
+	if err != nil {
+		return err
+	}
+
+	t.tmxs[url] = TMXResource{Level: lvl, url: url}
+	return nil
+}
+
+// Unload removes the preloaded level from the cache
```
otraore: +1 for documentation

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+func (t *tmxLoader) Unload(url string) error {
+	delete(t.tmxs, url)
+	return nil
+}
+
+// Resource retrieves the preloaded level, passed as a 'TMXResource'
```
otraore: > // Resource retrieves the preloaded level, passed as a 'TMXResource'
 
to > // Resource retrieves and returns the preloaded level of type 'TMXResource'

common/tmx_level.go
```diff
@@ -87,15 +84,10 @@ func (t ByFirstgid) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
...
 	tlvl := &TMXLevel{}
 	lvl := &Level{}
 
-	tmx, err := readTmx(r.url)
-	if err != nil {
-		return lvl, err
-	}
-
 	if err := xml.Unmarshal([]byte(tmx), &tlvl); err != nil {
 		fmt.Printf("Error unmarshalling XML: %v", err)
```
otraore: Shouldn't you also return this error?

common/tmx_level.go
```diff
@@ -139,7 +131,13 @@ func createLevelFromTmx(r Resource) (*Level, error) {
 
 	// Load in the images needed for the tilesets
 	for k, ts := range tlvl.Tilesets {
-		ts.Image = Files.Image(path.Base(ts.ImageSrc.Source))
+		url := ts.ImageSrc.Source
+		if err := engo.Files.Load("maps/" + url); err != nil {
```
otraore: I don't think this should be a hardcoded string

common/tmx_level.go
```diff
@@ -139,7 +131,13 @@ func createLevelFromTmx(r Resource) (*Level, error) {
...
+			return lvl, err
+		}
+		image, _ := engo.Files.Resource("maps/" + url)
```
otraore: Same as above

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+	url   string
+}
+
+func (r TMXResource) URL() string {
+	return r.url
+}
+
+// tmxLoader is responsible for managing '.tmx' files within 'engo.Files'
+type tmxLoader struct {
+	tmxs map[string]TMXResource
```
otraore: Better name? Also internal doc would be nice.

common/level.go
```diff
@@ -1,11 +1,8 @@
-//+build ignore
-
 package common
 
-import "engo.io/engo"
-
 import (
 	"engo.io/gl"
+        "engo.io/engo"
```
paked: Did you gofmt, this spacing doesn't look right?

common/tmx_filetype.go
```diff
@@ -0,0 +1,60 @@
...
+func (r TMXResource) URL() string {
+	return r.url
+}
+
+// tmxLoader is responsible for managing '.tmx' files within 'engo.Files'
+type tmxLoader struct {
+	levels map[string]TMXResource
+}
+
+// Load will load the tmx file and any other image resources that are needed
```
paked: I'd like a note here on how to generate TMX files. Probably something like:

> You can generate a TMX file with the Tiled map editor.

Newbrict: I really don't think this is necessary... googling tmx shows tiled map editor, it's the tiled map editor format, no one else uses it.

common/tmx_level.go
Newbrict: the error is returned

common/tmx_filetype.go
Newbrict: berry beast did it anyway

common/tmx_level.go
EtienneBruines: But it shouldn't log / print if it returns an error. The receiving end can parse / print / panic the error has he pleases.

common/tmx_level.go
```diff
@@ -170,17 +169,20 @@ func createLevelFromTmx(r Resource) (*Level, error) {
 	}
 
 	for i := 0; i < len(tlvl.ImgLayers); i++ {
-		curImg := Files.Image(path.Base(tlvl.ImgLayers[i].ImgSrc.Source))
+		curImg, err := PreloadedSpriteSingle(path.Base(tlvl.ImgLayers[i].ImgSrc.Source))
+		if err != nil {
+			return lvl, nil
```
EtienneBruines: This doesn't seem right?

common/tmx_level.go
```diff
@@ -201,18 +203,10 @@ func pointStringToLines(str string, xOff, yOff float64) []*Line {
 
 		p1 := engo.Point{x1, y1}
 		p2 := engo.Point{x2, y2}
-		newLine := &Line{p1, p2}
+		newLine := &engo.Line{p1, p2}
 
 		lines = append(lines, newLine)
```
EtienneBruines: This isn't correct either. `lines` is already of length `len(floatPts)-1`. Since you're looping over `i`, you should use `lines[i] = newLine`. Using `append` will only grow `lines` even more.

common/tmx_level.go
```diff
@@ -139,7 +132,13 @@ func createLevelFromTmx(r Resource) (*Level, error) {
 
 	// Load in the images needed for the tilesets
 	for k, ts := range tlvl.Tilesets {
-		ts.Image = Files.Image(path.Base(ts.ImageSrc.Source))
+		url := path.Join(path.Dir(tmxUrl), ts.ImageSrc.Source)
+		if err := engo.Files.Load(url); err != nil {
+			return lvl, err
+		}
+		image, _ := engo.Files.Resource(url)
+		texResource := image.(TextureResource)
```
EtienneBruines: This will `panic` on failure. Better use

```go
texResource, ok := image.(TextureResource)
if !ok {
    return nil, fmt.Sprintf("resource is not of type `TextureResource`: %q", url)
}
```

common/tmx_level.go
```diff
@@ -87,17 +84,13 @@ func (t ByFirstgid) Swap(i, j int)      { t[i], t[j] = t[j], t[i] }
...
 	lvl := &Level{}
 
-	tmx, err := readTmx(r.url)
-	if err != nil {
-		return lvl, err
-	}
-
 	if err := xml.Unmarshal([]byte(tmx), &tlvl); err != nil {
 		fmt.Printf("Error unmarshalling XML: %v", err)
+		return lvl, err
```
EtienneBruines: Idiomatic would be to `return nil, err` - consumes less resources and indicates that whatever `*Level` is being returned on failure, is useless.

common/tmx_level.go
```diff
@@ -139,7 +132,13 @@ func createLevelFromTmx(r Resource) (*Level, error) {
 
 	// Load in the images needed for the tilesets
 	for k, ts := range tlvl.Tilesets {
-		ts.Image = Files.Image(path.Base(ts.ImageSrc.Source))
+		url := path.Join(path.Dir(tmxUrl), ts.ImageSrc.Source)
+		if err := engo.Files.Load(url); err != nil {
+			return lvl, err
+		}
+		image, _ := engo.Files.Resource(url)
```
EtienneBruines: You should check this error as well, and return it on failure.

common/tmx_filetype.go
```diff
@@ -0,0 +1,62 @@
...
+func (t *tmxLoader) Load(url string, data io.Reader) error {
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(data)
+	if err != nil {
+		return err
+	}
+
+	lvl, err := createLevelFromTmx(buf.String(), url)
```
EtienneBruines: You're converting `[]byte` to `string` here, allocating new memory. 

However, when using it, you're casting it from `string` to `[]byte` again. You should simply pass along the `[]byte`.

common/level.go
```diff
@@ -57,17 +54,28 @@ type layer struct {
...
+			y := float32(i/int(setWidth)) * th
+
+			invTexWidth := 1.0 / float32(sheet.Image.Width)
+			invTexHeight := 1.0 / float32(sheet.Image.Height)
+
+			u := float32(x) * invTexWidth
+			v := float32(y) * invTexHeight
+			u2 := float32(x+tw) * invTexWidth
+			v2 := float32(y+th) * invTexHeight
+			t.Image = &Texture{id: sheet.Image.Texture, width: tw, height: th, viewport: engo.AABB{engo.Point{u, v}, engo.Point{u2, v2}}}
```
EtienneBruines: You're not setting the `width` and `height` for `t`, nor the `engo.Point`. 

Are you using the anonymous `engo.Point` defined in line 39?

common/tmx_filetype.go
```diff
@@ -0,0 +1,62 @@
+package common
+
+import (
+	"bytes"
+	"engo.io/engo"
```
EtienneBruines: Since I've already made enough comments to fill another commit: every package not within the standard library goes within its own group. 

So:

```go
import (
     "bytes"
     "fmt"
    "io"

    "engo.io/engo"
)
```

The tool `goimports` might help with that (that one splits the two groups, as is recommended by the Go team)). That basically does a `gofmt` with an organization of the `import` statement afterwards.

common/tmx_filetype.go
```diff
@@ -0,0 +1,62 @@
...
+	levels map[string]TMXResource
+}
+
+// Load will load the tmx file and any other image resources that are needed
+func (t *tmxLoader) Load(url string, data io.Reader) error {
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(data)
+	if err != nil {
+		return err
+	}
```
EtienneBruines: While I'm perfectly fine with this, since you're only using the `[]byte` representation, this might be of use:

```go
b, err := ioutil.ReadAll(data)
if err != nil {
    return err
}

// and now `b` is of type `[]byte`
```

common/tmx_level.go
EtienneBruines: Lines 105 and 113 should be fixed accordingly. (and also according to https://github.com/EngoEngine/engo/pull/315/files#r63825301 )

Comment on lines 120 through 125 (now we're at it):

```go
err = binary.Read(zlr, binary.LittleEndian, &nextInt)
if err != nil {
    // this is generally EOF
    break
}
```

I like the part where it's *generally* EOF, but let's check to be sure? If EOF, then break, else return error.

common/level.go
```diff
@@ -27,26 +24,26 @@ func (t *tile) Width() float32 {
 }
 
 func (t *tile) Texture() *gl.Texture {
-	return t.Image.Texture()
+	return t.Image.id
 }
 
 func (t *tile) Close() {
 	// noop
```
EtienneBruines: You could call `Close` on every `TextureResource` within every `*tile`.

common/level.go
```diff
@@ -27,26 +24,26 @@ func (t *tile) Width() float32 {
...
+	engo.Point
 	height int
 	width  int
-	Image  *Region
+	Image  *Texture
 }
 
 type tilesheet struct {
-	Image    *Texture
+	Image    *TextureResource
```
EtienneBruines: You can also use `Texture` instead; since you're not using anything `TextureResource` specific (and `TextureResource` is a wrapper for `Texture`). 

It looks like we're not using `tilesheet` anyway? Nor are we using `createTileset`? (nor could anyone else because they're unexported? Correct me if I'm wrong here.

EtienneBruines:
(Added the `do-not-merge` label in case someone accidentally thinks it's ready for merge at this state. As soon as comments are fixed, I'll remove the tag :smile: )

Something that doesn't have to be part of this PR, but might be something to think about: we could create a directory called `testdata`, which includes some kind of `.tmx` file, allowing us to create a `_test.go` file to test whether or not the loading actually works.

common/tmx_filetype.go
EtienneBruines: > it's the tiled map editor format, no one else uses it.

I think @Newbrict and @TheBerryBeast are right in this case, @paked. It's only to be used with the `tmx` format, those that won't be loading `tmx` files won't be using it anyway. This is something to note within some kind of `demo` (which doesn't exist yet? or is that the broken `tilemap` demo? ) - but doesn't really *have* to be here? (could be here, though)

common/level.go
TheBerryBeast: @EtienneBruines Both Texture and TextureResource contain a gl.Texture, but Texture has an additional viewport member along with function implementations for Drawable.  This functionality is not needed in this case as the tilesheet itself will not be rendered.  Let me know if I should use texture anyway.  Tilesheet is used by tmx_level.go along with createTileset.

EtienneBruines: Oh wow, my mistake!

common/level.go
TheBerryBeast: @EtienneBruines These members are set later on in createLevelTiles.

otraore:
It would also be nice if we could try to fix the tilemap demo (https://github.com/EngoEngine/engo/blob/master/demos/tilemap/tilemap.go) in this PR

TheBerryBeast:
@otraore I will take a look at it.

demos/tilemap/tilemap.go
```diff
@@ -1,54 +1,96 @@
...
 
+func (game *GameWorld) Exit()        {}
+func (game *GameWorld) Hide()        {}
+func (game *GameWorld) Show()        {}
+func (game *GameWorld) Type() string { return "GameWorld" }
+
 func main() {
-	World = &GameWorld{}
-	engo.Run("Hello", 1024, 640, false, World)
+	engo.Run(engo.RunOptions{Title: "TileMap Demo", Width: 800, Height: 800, ScaleOnResize: false}, &GameWorld{})
```
otraore: One last thing is to try to get this bit to match with the other demos. For example: https://github.com/EngoEngine/engo/blob/master/demos/hello-world/hello.go#L62

otraore:
Along with https://github.com/EngoEngine/engo/pull/315#discussion_r64124363 Each demo also has a [README](https://github.com/EngoEngine/engo/tree/master/demos/exit)  at its directory root. It would be nice if you add one in for the tilemap demo. Once these two things are fixed, this PR LGTM. I look forward to getting it merged.

PR #320, opened 2016-05-21
Added camera system EntityScroller and additional demo

TheBerryBeast:


demos/entityscroller/entityscroller.go
```diff
@@ -0,0 +1,159 @@
...
+	w.AddSystem(&common.EntityScroller{SpaceComponent: &character.SpaceComponent, TrackingBounds: levelData.Bounds()})
+	engo.Input.RegisterButton("moveup", engo.ArrowUp)
+	engo.Input.RegisterButton("moveleft", engo.ArrowLeft)
+	engo.Input.RegisterButton("moveright", engo.ArrowRight)
+	engo.Input.RegisterButton("movedown", engo.ArrowDown)
+}
+
+func (game *GameWorld) Exit()        {}
+func (game *GameWorld) Hide()        {}
+func (game *GameWorld) Show()        {}
```
EtienneBruines: These three methods are optional, and should be left out.

(if any other demo still has them, we should remove them from those as well)

demos/entityscroller/entityscroller.go
```diff
@@ -0,0 +1,159 @@
...
+type ControlSystem struct {
+	entity *Character
+}
+
+func (c *ControlSystem) Add(char *Character) {
+	c.entity = char
+}
+
+func (c *ControlSystem) Remove(ecs.BasicEntity) {
+	c.entity = nil
```
EtienneBruines: This is not good practice, at all. 

Note that `Remove` will get called as soon as *any* entity gets removed from *any* system. Even though it'll work in this scenario (because nothing is ever removed), this is not be what we should teach others. 

```go
func (c *ControlSystem) Remove(basic ecs.BasicEntity) {
	if c.entity != nil && basic.ID() == c.entity.ID() {
        c.entity = nil
    }
}
```

This would work, though. (given it compiles, haven't checked syntax)

demos/entityscroller/README.md
```diff
@@ -0,0 +1,11 @@
+# EntityScroller Demo
+
+## This demo is built upon the TileMap Demo
```
paked: Remove the `##`

otraore:
LGTM and it LGT @paked so merging it in.

PR #321, opened 2016-05-21
Slimmer engo clock internals

nitya-sattva:
- Short:
  * Faster tick method
  * Decrease structure size
  * Remove import of math package

- No more modulus on a float value, messy idea any way.

- No excessive divisions or multiplications inside Tick() or Time(), the Time internals are not optimal for this use case. (See links**)

- Size of the struct is now 40 bytes vs 72 bytes on the original, Time stores a extra pointer and caused padding inside the Clock struct. (Leaves room for 3 event pointers inside the 64 byte bounds of a cash line)

**Time.Sub()
https://golang.org/src/time/time.go?s=18822:18856#L618
**Time.UnixNano()
https://golang.org/src/time/time.go?s=24254:24284#L825
**Duration.Seconds()
https://golang.org/src/time/time.go?s=17832:17867#L579

***Note: In case Delta() is only called once or twice every frame this is fine, but when Delta() is called often it might be preferable to move the conversion to seconds back in to Tick() and sacrifice some memory for it.

### DIFF
--- a/clock.go
+++ b/clock.go
@@ -2,58 +2,62 @@ package engo
 
 import (
 	"time"
-
-	"github.com/luxengine/math"
 )
+
+// The amound of nano seconds in a second.
+const secondsInNano int64 = 1000000000
 
 // A Clock is a measurement built in `engo` to measure the actual frames per seconds (framerate).
 type Clock struct {
-	elapsed float32
-	delta   float32
-	fps     float32
-	frames  uint64
-	start   time.Time
-	frame   time.Time
+	counter   uint32
+	perSecond uint32
+
+	deltaStamp int64
+	elapsStamp int64
+	frameStamp int64
+	startStamp int64
 }
 
 // NewClock creates a new timer which allows you to measure ticks per seconds. Be sure to call `Tick()` whenever you
 // want a tick to occur - it does not automatically tick each frame.
 func NewClock() *Clock {
+	currStamp := time.Now().UnixNano()
+
 	clock := new(Clock)
-	clock.start = time.Now()
-	clock.Tick()
+	clock.frameStamp = currStamp
+	clock.startStamp = currStamp
 	return clock
 }
 
 // Tick indicates a new tick/frame has occurred.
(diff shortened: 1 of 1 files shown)


clock.go
```diff
@@ -2,58 +2,62 @@ package engo
...
 	}
 }
 
 // Delta is the amount of seconds between the last tick and the one before that
 func (c *Clock) Delta() float32 {
-	return c.delta
+	return float32(float64(c.deltaStamp) / float64(secondsInNano))
 }
 
 // FPS is the amount of frames per second, computed every time a tick occurs at least a second after the previous update
```
EtienneBruines: This should also become `Fps`, then.

nitya-sattva:
My bad, I missed the name change when moving code from my project to the engo repo.
@EtienneBruines Did you want the name change or the original ?

EtienneBruines:
`Fps` is idiomatic, so both should be `Fps` (good catch! :smile: )

Thank you for the PR, the rest all LGTM. Awesome!

clock.go
```diff
@@ -2,58 +2,62 @@ package engo
...
+		c.elapsStamp %= secondsInNano
+		c.perSecond = c.counter
+		c.counter = 0
 	}
 }
 
 // Delta is the amount of seconds between the last tick and the one before that
 func (c *Clock) Delta() float32 {
-	return c.delta
+	return float32(float64(c.deltaStamp) / float64(secondsInNano))
```
EtienneBruines: Even though a `float64` divided by a `float64` is faster than using `float32` (because of some weird optimizaiton in the Go library), this might not be the case when you have to cast it to `float32` again? 

Not sure though, but just a thought.

nitya-sattva:
I think the speed is about the same on a 64bit machine, when running as 32bit it might be the case. Tbh my mind was on the precision when casting the total of an int64 to a float32.

EtienneBruines:
Thank you very much @nitya-sattva for your contribution! It is much appreciated :smile:

PR #323, opened 2016-05-23
Removed unneeded file from pong demo

otraore:


EtienneBruines:
LGTM, merge when testing is done. :smile:

PR #327, opened 2016-05-23
Fix: Delta Time in the first two frames

nitya-sattva:
By doing the start tick before the main loop the frame stamp inside the clock gets reset and this will minimize the delta time on the first frame.

Moving the Tick inside loop to run first reduces the delta time on the first frame even more and ensures a normal delta time in the second frame. (Having the tick first is also a plus if fixed physics are added at some point)

Note: Reordered the main loop to be the same on all targets, there is no reason to have them update in a different order ? The web build still seems to build and run fine, I'm not able to test the mobile build.

---

Other options considered:

- Move the clock creation to the start of the loop ? The clock is public and can be accessed in the loading before the main loop starts. Having 'nil' dangling out of the public interface seems bad design.

- Adding a reset option ? It would almost be the same and works as well but tempting people to call a function like reset ...


paked:
LGTM

otraore:
LGTM as well.

PR #330, opened 2016-05-23
Font atlas

### LINKED ISSUE #232: Font atlas
A lot of our font rendering is being done on the CPU-side (read: all of it). 

Instead, we should render each character (pretty much the ASCII alphabet) into a single "texture", and use an "atlas" (sort of like a spritesheet), to indicate which character OpenGL should draw. 

A video on how to do this in Java: https://www.youtube.com/watch?v=mnIQEQoHHCU&list=PLRIWtICgwaX0u7Rf9zkZhLoLuZVfUksDP&index=32

### LINKED ISSUE #308: Font render newlines
Currently, whenever the text a font needs to render (`myFont.Render("text\r\n")`) contains some kind of newline (`\r`, `\n`, or `\r\n` (I don't care which)), it gets ignored. 

It'd be nice to support multiline strings like this. At least some simple approach.

### LINKED ISSUE #317: Font partially transparency colors
# Given:
```go
	fnt := common.Font{
		URL:  "fonts/Roboto-Regular.ttf",
		FG:   color.Black,
		BG:   color.RGBA{255, 255, 255, 128},
		Size: 32,
	}
```

# Expected:
Partially transparent background. When the things behind that entity are white, then it should still be white. 

# Actual:

As if there's "black" behind the image the font is rendering. Everything behind it is white, so it should render 100% white background.

EtienneBruines:
This is done for now, but still needs review before merging. 

Fixes #232. Fixes #308. Fixes #317. 

The result? `LetterSpacing` and `LineSpacing` are also supported :smile: 

## Known bugs

Using (`LetterSpacing` or `LineSpacing`) in combination with `font.BG`, results in something a bit weird (the spacing isn't affected by the background):

Should be fixed, but don't really feel like doing it in this PR (as I'm not yet sure how).


common/font.go
```diff
@@ -145,3 +148,152 @@ func (f *Font) Render(text string) Texture {
...
+			fallthrough
+		case char == 'Q':
+			fmt.Println(string(char), float32(vm.TopSideBearing), float32(g.Bounds.Min.X), float32(g.Bounds.Min.Y), float32(g.Bounds.Max.X), float32(g.Bounds.Max.Y), float32(g.Bounds.Max.Y-g.Bounds.Min.Y))
+		}
+
+		yB := vm.TopSideBearing
+		//yB := g.Bounds.Max.Y - g.Bounds.Min.Y
+		if yB > maxYBearing {
+			maxYBearing = yB
+		}
```
EtienneBruines: The problem currently is, that the `vm.TopSideBearing` and `g.Bounds.Max.Y` and `g.Bounds.Min.Y` need to be combined to result in the total height. 

Maybe `vm.TopSideBearing - g.Bounds.Min.Y`? (`Min.Y` is negative)

common/font.go
```diff
@@ -145,3 +146,154 @@ func (f *Font) Render(text string) Texture {
...
+		totalHeight fixed.Int26_6
+
+		fupe        = fixed.Int26_6(f.TTF.FUnitsPerEm())
+		maxYBearing = fixed.Int26_6(0)
+
+		currentX float32
+		maxX     float32
+
+		totalString string // TODO; string is immutable, so this is relatively inefficient; see also where we use `totalString += string(char)`
+		subString   string
```
EtienneBruines: If anyone has a good idea to make this more efficient, be my guest :wink:

common/font.go
```diff
@@ -145,3 +146,154 @@ func (f *Font) Render(text string) Texture {
...
+
+		drawCurY int
+	)
+
+	// The "full image"
+	nrgba := image.NewNRGBA(image.Rect(0, 0, 1200, 5000)) // way too big; hopefully
```
EtienneBruines: AFAIK, we can't resize an `image.NRGBA` to whatever size we want to, and we need it to be big enough to fit everything in there. We can't "precompute" the size efficiently, either.

common/font.go
```diff
@@ -145,3 +146,154 @@ func (f *Font) Render(text string) Texture {
...
+		g := truetype.GlyphBuf{}
+		err := g.Load(f.TTF, fupe, idx, font.HintingNone)
+		if err != nil {
+			log.Println("Error creating font atlas:", err)
+			return atlas
+		}
+
+		int26Width += hm.AdvanceWidth
+
+		//atlas.Width[char] = float32(hm.AdvanceWidth * fixed.Int26_6(f.Size) / fupe)
```
EtienneBruines: Should be removed.

common/font.go
```diff
@@ -145,3 +146,154 @@ func (f *Font) Render(text string) Texture {
...
+		atlas.Width[char] = float32(g.AdvanceWidth * fixed.Int26_6(f.Size) / fupe)
+
+		currentX = float32(int26Width * fixed.Int26_6(f.Size) / fupe)
+		if currentX > maxX {
+			maxX = currentX
+		}
+
+		//yB := vm.TopSideBearing
```
EtienneBruines: Should be removed (same for two lines below)

common/font.go
```diff
@@ -145,3 +146,154 @@ func (f *Font) Render(text string) Texture {
...
+	draw.Draw(nrgba, image.Rect(0, drawCurY, subimg.Bounds().Max.X, drawCurY+subimg.Bounds().Max.Y), subimg, image.ZP, draw.Src)
+	drawCurY += subimg.Bounds().Max.Y
+
+	int26Height += maxYBearing
+	totalHeight += maxYBearing
+	maxYBearing = fixed.Int26_6(0)
+
+	subString = ""
+	int26Width = 0
+	currentX = 0
```
EtienneBruines: These three lines also make no sense here (even though they make sense within the loop).

common/render.go
```diff
@@ -118,6 +118,42 @@ func (rs *RenderSystem) New(w *ecs.World) {
...
+			render.shader = LegacyHUDShader
+		case ComplexTriangles:
+			render.shader = LegacyHUDShader
+		case Text:
+			render.shader = TextHUDShader
+		default:
+			render.shader = HUDShader
+		}
+	}
+
```
EtienneBruines: These set default shaders for those that have no shaders set, depending on the `Drawable`-type, so you don't explicitly have to say which shader you want to use, when you're using shapes / text

common/render_shaders.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/luxengine/math"
 )
 
-const bufferSize = 10000
+// UnicodeCap is the amount of unicode characters the fonts will be able to use, starting from index 0.
+var UnicodeCap = 200
+
+var bufferSize = 10000
```
EtienneBruines: No idea why I changed it to `var`, should change back.

common/render_shaders.go
```diff
@@ -783,6 +775,320 @@ func (l *legacyShader) Post() {
...
+
+	if l.cameraEnabled {
+		for _, system := range w.Systems() {
+			switch sys := system.(type) {
+			case *CameraSystem:
+				l.camera = sys
+			}
+		}
+		if l.camera == nil {
+			log.Println("WARNING: BasicShader has CameraEnabled, but CameraSystem was not found")
```
EtienneBruines: This one is called `TextShader`, not `BasicShader`.

common/render_shaders.go
```diff
@@ -783,6 +775,320 @@ func (l *legacyShader) Post() {
...
+		setBufferValue(buffer, 16+offset, currentY+atlas.Height[char], &changed)
+		setBufferValue(buffer, 17+offset, atlas.XLocation[char]/atlas.TotalWidth, &changed)
+		setBufferValue(buffer, 18+offset, (atlas.YLocation[char]+atlas.Height[char])/atlas.TotalHeight, &changed)
+		setBufferValue(buffer, 19+offset, tint, &changed)
+
+		currentX += modifier * (atlas.Width[char] + float32(txt.Font.Size)*txt.LetterSpacing)
+	}
+
+	return changed
+}
```
EtienneBruines: This is the "most important" part of the shader: it tells OpenGL for each corner, where it should go. The first two (out of 5), give the location relative to the origin of the object. The two after that, give the location relative to the FontAtlas (value between 0 and 1 to indicate where on that FontAtlas image the character is drawn). The last one gives the `tint` / color-filter.

common/render_shaders.go
```diff
@@ -783,6 +775,320 @@ func (l *legacyShader) Post() {
...
+	txt, ok := ren.Drawable.(Text)
+	if !ok {
+		unsupportedType(ren.Drawable)
+	}
+
+	atlas, ok := atlasCache[txt.Font.URL]
+	if !ok {
+		// Generate texture first
+		atlas = txt.Font.generateFontAtlas(UnicodeCap)
+		fmt.Println("Generating because !ok 1050")
```
EtienneBruines: This line should be removed.

demos/shapes/README.md
```diff
@@ -0,0 +1,28 @@
...
+Finally, we add it to the `RenderSystem`:
+
+```go
+for _, system := range w.Systems() {
+    switch sys := system.(type) {
+    case *common.RenderSystem:
+        sys.Add(&triangle1.BasicEntity, &triangle1.RenderComponent, &triangle1.SpaceComponent)
+    }
+}
+```
```
EtienneBruines: This file apparently didn't have a `README` yet, so now it does :smile:

engo_glfw.go
```diff
@@ -326,6 +326,7 @@ func SetVSync(enabled bool) {
 func init() {
 	runtime.LockOSThread()
 
+	Grave = Key(glfw.KeyGraveAccent)
```
EtienneBruines: Not really part of this PR, but needed it for TrafficManager :laughing:

demos/text/text.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	label1 := MyLabel{BasicEntity: ecs.NewBasic()}
+	label1.RenderComponent.Drawable = common.Text{
+		Font: fnt,
+		Text: "Hello world !",
+	}
+	label1.SetShader(common.HUDShader)
+
+	common.UnicodeCap = 1900
```
EtienneBruines: This should be removed (was needed to test for arabic, which works).

common/render_shaders.go
```diff
@@ -783,6 +775,320 @@ func (l *legacyShader) Post() {
...
+
+	var currentX float32
+	var currentY float32
+
+	var modifier float32 = 1
+	if txt.RightToLeft {
+		modifier = -1
+	}
+
+	for index, char := range txt.Text {
```
EtienneBruines: TODO: we could cache here (for this specific string), so we don't have to compute this again.

common/render_shaders.go
```diff
@@ -13,9 +13,9 @@ import (
 )
 
 // UnicodeCap is the amount of unicode characters the fonts will be able to use, starting from index 0.
-var UnicodeCap = 200
+const UnicodeCap = 200
```
EtienneBruines: This one shouldn't be `const`, because users should be able to edit this (i.e. it needs to be set to 1900 for arabic, but because of the performance cost, it's defaulted to 200)

EtienneBruines:
I think using the tables at https://golang.org/pkg/unicode/ might be better than simply computing the first 200 chars? Letting people define lists like that, with `Latin` being the default? Thoughts?

paked:
Whoever merges this in could they please make an issue regarding the background color bugs.

>  ## Known bugs
> 
> Using (`LetterSpacing` or `LineSpacing`) in combination with `font.BG`, results in something a bit weird (the spacing isn't affected by the background):
> 
>

Issue #313, opened 2016-05-18
engo.Point{}.Within 


EtienneBruines:
@otraore thought it'd be nice to be able to say

```go
point := engo.Point{1, 2}
spaceComponent := common.SpaceComponent{
    Position: engo.Point{3, 4},
    Width: 5,
    Height: 6,
}

// Specifically this line:
point.Within(spaceComponent)

// This you can already do, and should result in exactly the same:
spaceComponent.Within(point)
```

mewmew:
How do we make `Point` not depend on the `common` package? Through an interface?

If we define a method `Within` on point, it would make sense to rename the `SpaceComponent.Within` method to `Contains`. Perhaps that's just a personal preference. What do you think?

EtienneBruines:
> How do we make Point not depend on the common package? Through an interface?

You're right, we can't. (interface doesn't really make it any better, at least not for a "convenience" method, imho)

`Contains` SGTM. Awaiting response from others :smile:

otraore:
> https://github.com/EngoEngine/engo/issues/313#issuecomment-220161908
> That’s the rename we were looking for. @mewmew
> Contains does make a lot more sence for a space component
> And Within more sense for a Point
> @otraore in the gitter chat

EtienneBruines:
> Harrison Shoebridge @paked 23:27
> Yep!
> That's what I was trying to say

EtienneBruines:
Up for grabs.

paked:
Closes by #333

otraore:
I disagree @paked because https://github.com/EngoEngine/engo/pull/333 only fixes one of the things mentioned in this issue. The other one `Poing.Within` is still up for grabs. It's a tough issue.

mewmew:
> > How do we make Point not depend on the common package? Through an interface?
>
> You're right, we can't. (interface doesn't really make it any better, at least not for a "convenience" method, IMHO)

> The other one Poing.Within is still up for grabs. It's a tough issue.

@otraore, could you elaborate? I thought we were not going to add a `Within` method to `Point`, as that would require it to have knowledge about `common`. Or are we aiming to introduce some form of interface to encapsulate the notion of an entity in which another entity may be contained within?

One option may be.

```go
package engo

// A Container is a 2-dimensional closed shape which may contain a set of points.
type Container interface {
   // Contains reports whether the container contains the given point.
   Contains(p Point) bool
}

// A Point is an X, Y coordinate pair. The axes increase right and down.
[...]

PR #333, opened 2016-05-26
Renamed SpaceComponent.Within to Contains, #313

akiross:
My first PR (ever, probably) so, please, check everything is fine. I ran `go test engo.io/engo` and `go test engo.io/engo/common` and they pass; hoping that's enough.

This should fix the issue in #313 regarding the SpaceComponent.Within renaming to Contains.

### DIFF
--- a/common/collision.go
+++ b/common/collision.go
@@ -87,9 +87,9 @@ func (sc SpaceComponent) Corners() (points [4]engo.Point) {
 	return
 }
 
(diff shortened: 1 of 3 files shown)


paked:
LGTM if compiles.

Thanks for the PR @akiross. It really means a lot to us :D

paked:
@akiross 🎉

PR #338, opened 2016-05-28
Tilemap update

otraore:


demos/tilemap/tilemap.go
```diff
@@ -90,10 +90,9 @@ func (game *GameWorld) Type() string { return "GameWorld" }
 
 func main() {
 	opts := engo.RunOptions{
-		Title:         "TileMap Demo",
-		Width:         800,
-		Height:        800,
-		ScaleOnResize: false,
```
otraore: No real need to set this to false. It looks fine when resized and even though it loses quality slightly it's better than white ends on each sides.

paked:
🎉

PR #343, opened 2016-05-31
Fix Input behavior for web builds

### LINKED ISSUE #342: Input Demo - This is broken on web builds
When running the input demo in the browser this almost never fires: btn.JustPressed()

Looked over the code and I think this is because the callback to 'Input.keys.Set()' might happen before 'Input.Update()' and the code expects it to be set after. (Did not verify, but seems like the only logical conclusion)

The only way around this that I can see is to create a temp buffer to store key state from the callbacks and then pass them on to 'Input.keys.Set()' after 'Input.Update()' ?

---

Edit: Buffer fix works, demo code is below. I can move it in to a pull-able branch if it is the best solution. (Branch it is in now is dirty)
- Input Demo: Better for testing
  https://github.com/nitya-dev/engo/commit/37248588d5254ab19103ec572fc741927253ca23
- Buffer keys in web build: File 'engo_js.go'
  https://github.com/nitya-dev/engo/commit/ea3b15843a60c1e050da8a9718601a6b78b7d62e#diff-a2ad628a9a250d105a8a9df7b5fe7698

nitya-dev:
Adds a web key buffer to the netgo build to deal with the async nature of web browsers and JavaScript. This resolves #342 in a simple way whiteout the need for massive changes.

Note: The resetting of the buffer leaves memory referenced that could leak when used without care. Sins this buffer is alive as long as the application is alive and the size is limited to the max amount of key events in +-16ms  this seems like a valid use case to avoid garbage generation.

### DIFF
--- a/demos/input/input.go
+++ b/demos/input/input.go
@@ -13,10 +13,18 @@ type DefaultScene struct{}
 func (*DefaultScene) Preload() {}
 func (*DefaultScene) Setup(w *ecs.World) {
 	w.AddSystem(&common.RenderSystem{})
+
+	// Register the input system responding
+	// to the input actions registered below.
 	w.AddSystem(&InputSystem{})
 
-	engo.Input.RegisterAxis("sideways", engo.AxisKeyPair{engo.A, engo.D})
+	// Register a button that is triggered when
+	// the state of 'Space' or 'Enter' is changed.
 	engo.Input.RegisterButton("action", engo.Space, engo.Enter)
+
+	// Register an axis where 'A' will return a
+	// negative value and 'D' returns a positive value.
+	engo.Input.RegisterAxis("sideways", engo.AxisKeyPair{engo.A, engo.D})
 }
 
 func (*DefaultScene) Type() string { return "Game" }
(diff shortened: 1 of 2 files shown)


paked:
@nitya-dev What's the status of this? Mergeable?

Either way, it'd be great to get it back up to date with master.

nitya-dev:
It was 'mergeable', I will merge master in to the branch tonight.

Is is a bit of a hack but it does solve the issue and I did not find a other way to solve it. (aka I'm using the same work around in the other input system.)

demos/input/input.go
```diff
@@ -52,8 +52,16 @@ func (c *InputSystem) Update(dt float32) {
 		fmt.Println(v)
 	}
 
+	if btn := engo.Input.Button("action"); btn.Down() {
```
otraore: Can't we assign `btn` once and reuse it?

demos/input/input.go
```diff
@@ -52,8 +52,16 @@ func (c *InputSystem) Update(dt float32) {
...
+		fmt.Println("Still down!")
+	}
+
 	if btn := engo.Input.Button("action"); btn.JustPressed() {
```
otraore: Because same thing over here

demos/input/input.go
```diff
@@ -52,8 +52,16 @@ func (c *InputSystem) Update(dt float32) {
...
-		fmt.Println("DOWN!")
+		fmt.Println("Key just pressed!")
+	}
+
+	if btn := engo.Input.Button("action"); btn.JustReleased() {
```
otraore: And here

nitya-dev:
Got refused on my end, reason:

The api to set the key state is public so one can not depend on the browser or glfw to be the only ones trying to set those states. This work around depends on this and does not fix the problem if others try to set key states in the middle of a frame in a mobile or desktop build.

On my end I moved the buffer in to the key manager and then update states before sending out events. If I find time I will setup a new pull request with that fix for engo.

PR #352, opened 2016-06-08
Fix EdgeScroller for bottom & right edges

laher:
Use engo.GameWidth(), engo.GameHeight() in place of engo.CanvasWidth(), engo.CanvasHeight().

This fixes the 'edgescroller' demo.


EtienneBruines:
Pff, this has been a while. 

IIRC, this one was "on hold" because of an issue with retina displays. Not sure if that's fixed already, though - and if so, in which PR this was done.

otraore:
It's also branched off from master @EtienneBruines

EtienneBruines:
Replaced by #367. Thank you @laher for the help you gave me with #367 a week ago.

Issue #353, opened 2016-06-08
Feature Request - Add layer information to TMX TileMaps tiles


randomandy:
Please correct me if I'm wrong, but:

After playing around with engo for a few days now, I get the feeling that it'd be necessary for TileMaps to contain their different layer information.
That way it's possible to assign each layer's tiles (or objects, images for that matter) a specific **ZIndex** (`tile.RenderComponent.SetZIndex(x)`)

Right now engo handles TMX maps by distinguishing between **Tiles** (`TMXResource.Level.Tiles`) and **Images** (`TMXResource.Level.Images`).
So if I'm not mistaken, currently one has to range over all Tiles and Images:

```
resource, err := engo.Files.Resource("example.tmx")
if err != nil {
    panic(err)
}

tmxResource := resource.(common.TMXResource)
levelData := tmxResource.Level
```

and then:

```
for _, v := range levelData.Tiles {
    if v.Image != nil {
        tile := &Tile{BasicEntity: ecs.NewBasic()}
        tile.RenderComponent = common.RenderComponent{
            Drawable: v,
            Scale:    engo.Point{1, 1},
        }
        tile.SpaceComponent = common.SpaceComponent{
            Position: v.Point,
            Width:    0,
            Height:   0,
        }
        tileComponents = append(tileComponents, tile)
    }
}

for _, v := range levelData.Images {
    if v.Image != nil {
        tile := &Tile{BasicEntity: ecs.NewBasic()}
        tile.RenderComponent = common.RenderComponent{
            Drawable: v,
            Scale:    engo.Point{1, 1},
        }

        tile.RenderComponent.SetZIndex(2)

        tile.SpaceComponent = common.SpaceComponent{
            Position: v.Point,
            Width:    0,
            Height:   0,
        }

        tileComponents = append(tileComponents, tile)
    }
}

```

From what I understand after playing around with [Tiled](http://www.mapeditor.org/), it would make sense to distinguish between Tiles, Objects and Images.
Therefor it would be necessary that each Tile would also "know" the Tile Layer it belongs to. Same goes for Objects and Images of course.

For instance when ranging over []*common.tile in your TMXResource.Level.Tiles, maybe each element could contain the name of its tile layer:

```
for _, tile := range levelData.Tiles {
    // tile.Layer could contain the information of the tile's layer
}
```

Or maybe even better, let common.TMXResource.Level contain an array of all layers (of all types: tile layer, object layer, image layer):

```
[...]

PR #355, opened 2016-06-08
Tile layer name

randomandy:
Added LayerName to common.tile struct (for Tiles and Images)

### DIFF
--- a/common/level.go
+++ b/common/level.go
@@ -49,7 +49,8 @@ func (t *tile) View() (float32, float32, float32, float32) {
 
 type tile struct {
 	engo.Point
-	Image *Texture
+	Image     *Texture
+	LayerName string
 }
 
 type tilesheet struct {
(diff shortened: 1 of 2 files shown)


randomandy:
Basic fix for #353

otraore:
It LGTM, I'm going to leave it open to see the thoughts of other people. I assume you also tested this out? But with that said It would be nice if you can add in support for all the data available in the tmx layer element to our [TMXLayer](https://github.com/EngoEngine/engo/blob/master/common/tmx_level.go#L35) object. Also maybe have a way to get the layer from the tile object rather than layer properties on the tile. . Eventually we want to support the [full tmx specification](http://doc.mapeditor.org/reference/tmx-map-format/) so this would help a lot.

randomandy:
Ok I'm gonna try to get the full specs done this weekend. If you want you can merge this one, and I'm gonna create a new PR when I'm done with the rest.

otraore:
@randomandy I would appreciate it if you at least covered the Layer spec in this PR since that's the one it's focused on. Don't need to do it for every element in this PR but at least for the Layer one. Thank you again, appreciate it.

randomandy:
I'm closing this PR for now as I'm adding the full feature set we discussed earlier and in #353 . Gonna open a new PR as soon as it's all done and tested

paked:
Awesome!

randomandy:
Done, please see #357 for details

PR #356, opened 2016-06-09
Add Within helper method to engo.Point

otraore:
As [discussed by @mewmew](https://github.com/EngoEngine/engo/issues/313#issuecomment-221848332) 
Fixes: #313

### DIFF
--- a/math.go
+++ b/math.go
@@ -16,7 +16,7 @@ type AABB struct {
 	Min, Max Point
 }
 
-// A Container is a 2-dimensional closed shape which may contain a set of points.
+// A Container is a 2D closed shape which contains a set of points.
 type Container interface {
 	// Contains reports whether the container contains the given point.
 	Contains(p Point) bool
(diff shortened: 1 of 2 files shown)


math.go
```diff
@@ -16,6 +16,12 @@ type AABB struct {
 	Min, Max Point
 }
 
+// A Container is a 2-dimensional closed shape which may contain a set of points.
```
paked: > A Container is a 2D shape which may or may not contain a given point.

paked: That would be my godoc comment

otraore: I disagree, the points mentioned means the container has points in it and the container encloses it. It's not talking about whether or not the container contains a specific point but that a container MUST have a set of point`s` otherwise it's not a container, it's just one point.

paked: @otraore OK I agree with this assessment.

paked:
Woo!

PR #364, opened 2016-06-16
Reorganize the input system

### LINKED ISSUE #342: Input Demo - This is broken on web builds

### LINKED ISSUE #349: Keys namespace
As of this moment, each key is under the same general namespace. So if you want the `w` key, you would get `engo.W`.

My proposition:
Namespace so we could get it with `engo.Keys.W`. This way we wouldn't "dirty" the general namespace and it would be easier with an IDE for example to understand what that `engo.W` means.

nitya-dev:
Wip: Looking for feedback

ToDo: Add docs back where needed
ToDo: Only the input demo is up to date
ToDo: Implement mouse movement axis
ToDo: Check if there is no obsolete code left

---

### Generalize the core input

Input to a game can be many things and has some specialized requirements depending on the game and/or the platform. As it stand now the input is a core part of engo and it is hard to modify or extend.

The changes in this pull request isolate the core of the input and then provide a safe and clean way to use it. By doing this there is a cleaner path to extend, swap out or setup a secondary input system when required.

*@Sendoushi mentioned this desire in gitter and it is used by me to push and pop input behavior depending on the game state, when attempting to do this in the old system it became messy due to the use of global variables that link it all up.*

### Minimizes string lookups

When looking at the benchmarks of the old system there is a big diffidence between the most optimal way of doing things and the worst way of doing things. This is mostly due to the string based lookups that happen every frame.

The new system ditches the string lookups in the game loop and moves them to the setup and loading section of the game. Once running all lookups are based on id's.

- This also adds the option to check and implement fall-backs for systems that depend on some inputs to be setup. The ability to do this in turn helps when adding saving/loading of user preferences *(See: The input demo)*

*Note: The optimal benchmarks where possible because the axes and buttons (ab)-used global variables. The new system no longer has an optimal or sub-optimal path, performance wise that benchmark would now sit between the two extremes*

### Other changes worth mentioning

- Cleans up the testing to make it easier to add edge case.

- Mouse buttons now act like key codes and can be used in the axis and button managers.

- Solves the out of order state changes that caused the web input to be broken. (Resolves #342)

- Fixes behavior on the axis where holding down both min and max would not return zero. (I assume zero is the valid outcome of this edge case)

- Extend axis with the option to check the state of an axis in the same way you can check state for codes and buttons. Plus the option to check state for a specific axis direction (min/max). *(See: The input demo)*

- Fixes a issue in buttons where just pressed/released would trigger multiple times when interacting with a secondary key while holding down a primary key.

- And it touches on the proposal made in #349 by moving the codes in to a separated package and using the 'Key' and 'Mouse' prefix. (Resolves #349)

- The way key and mouse codes are setup adds a safe and easy way to add custom codes to the action manager when needed. (Controllers, touche, ...)

*Edge cases mentioned above are added to the new testing*

### Things I played with but left out

- Act, Axis, Button Events: Hard to generalize, a generalized event system slows things down and a specialized system is to focused on one use case.

- Combo Axis/Button and Sequence managers: When trying to make these more general they slowed down dramatically. These managers all have there own unique sets of behaviors, witch set of behaviors you need depends on the game or application your making

*Leaving these out of the engine and adding some specialized demos on how to achieve these behaviors would be preferable to me.*


engo.go
```diff
@@ -5,14 +5,20 @@ import (
...
+	"engo.io/engo/act"
 )
 
 var (
 	// Time is the active FPS counter
 	Time *Clock
 
 	// Input handles all input: mouse, keyboard and touch
-	Input *InputManager
+	Input *InputMgr
```
Sendoushi: I think these `Mgr` should be `Manager`. More semantic and as such easier to understand.

engo.go
```diff
@@ -118,21 +124,35 @@ func Run(o RunOptions, defaultScene Scene) {
...
+	Buttons = Input.Buttons()
 
-		Input.RegisterAxis(DefaultHorizontalAxis, AxisKeyPair{A, D}, AxisKeyPair{ArrowLeft, ArrowRight})
-		Input.RegisterAxis(DefaultVerticalAxis, AxisKeyPair{W, S}, AxisKeyPair{ArrowUp, ArrowDown})
+	if opts.StandardInputs {
+		log.Println("Using standard inputs")
 
-		Input.RegisterAxis(DefaultMouseXAxis, NewAxisMouse(AxisMouseHori))
-		Input.RegisterAxis(DefaultMouseYAxis, NewAxisMouse(AxisMouseVert))
+		Axes.SetNamed(
```
Sendoushi: Semantically speaking... what is set named?

engo_js.go
```diff
@@ -200,6 +207,7 @@ func rafPolyfill() {
 // RunIteration runs one iteration per frame
 func RunIteration() {
 	Time.Tick()
+	Input.clear()
 	Input.update()
 	currentWorld.Update(Time.Delta())
```
Sendoushi: Unrelated: I think `render` shouldn't know about `currentWorld`. This could instead be a mailbox event.

engo.go
nitya-dev: Hmm, using consistent suffixes (like Mgr) is helpful for someone maintaining the code, using  Manager is annoying for a maintainer (More typing and longer, more complex lines to read). 

Using the full word is only helpful for people reading the code for the first time, learning the meaning of a consistent set of suffixes happens fast.

Sendoushi: Well, I don't agree. I'm against things like variable names as `o` or `s` because then I have to search for signatures or whatever. Even when I'm the guy who did the code. I always prefer things a little more verbose but easier to understand what they are.

Anyway, just my humble opinion!

nitya-dev: I agree that variable names need to be verbose, but I prefer using a short suffixes to add extra info on a object name.

engo_js.go
otraore: @Sendoushi Please explain what you mean by this in detail somewhere else (issue, gitter) because you keep mentioning it and I have no idea what you're saying.

Sendoushi: @otraore Made a pull request with a possible solution instead :)

otraore: Just saw and responding to it.

engo.go
paked: I'm not a fan of this sort of shortening. I'd strongly prefer having `InputManager`.

act/act-mgr.go
```diff
@@ -0,0 +1,113 @@
...
+
+const KeyCode = Code(1 << 16)
+const MouseCode = Code(2 << 16)
+
+const StateIdle = State(0)
+const StateActive = State(1)
+const StateJustIdle = State(2)
+const StateJustActive = State(3)
+
+////////////////
```
paked: This isn't idiomatic godoc commenting

act/act-mgr.go
```diff
@@ -0,0 +1,113 @@
+package act
```
paked: All this needs to be documneted.

input.go
```diff
@@ -1,68 +1,118 @@
...
+	// Mouse is InputMgr's reference to the mouse. It is recommended to use the
 	// Axis and Button system if at all possible.
 	Mouse Mouse
 
-	axes    map[string]Axis
-	buttons map[string]Button
-	keys    *KeyManager
+	actMgr    *act.ActMgr
+	axisMgr   *act.AxisMgr
+	buttonMgr *act.ButtonMgr
```
paked: We could just call these `buttons`, `axes` and `acts`?

act/act-mgr.go
nitya-dev: I disagree, from the go docs: 
> Comments that are not adjacent to a top-level declaration are omitted from godoc's output, with one notable exception.

So godoc allows the user to add top level comments that are ignored. If they are not ignored this is a bug or they need to adjust there documentation. Now I do understand you do not like this comment style, I remember that from the last pr and I was planning to remove it when it was ready to be merged.

nitya-dev:
I tried, I'm to old for this social media bullshit. I have 5.5 months left to finish the port of the two games to golang. If that is done I will setup a repo with MIT code you can backport. gl

nitya-dev:
This was my last push, if you want the changes I suggest you clone my repo and adjust/patch/doc it your self.  I now remember why I dislike the github culture and I will be removing my account again soon.

act/act-mgr.go
paked: Also the filename is rather redundant. Probably should be `manager.go`.

Noofbiz:
Outdated and there's a better input system suggestion in #349 and #433

PR #365, opened 2016-06-16
A set of propositions

Sendoushi:
After checking some things on the main code I found some things that I would change and this pull request tries to address some of those. Please don't accept it because I need to update the `demos` if this is to go through.

Changes:
- Remove function in Mailbox (untested)
- Common Mailbox for all (I think that having a global Mailbox will be easier to use and it helps with background stuff for example)
- `SceneSettings` instead of interface functions
- Removed mailbox from scene. Instead, it may use the global one.
- `InitGl` instead of using the global `func init()`
- Separate `Init` from `Run`
- Dispatch preparation and iteration messages (Render shouldn't know about scenes, worlds... It will help unpatternize the engine)
- Move `WindowResizeMessage` to the place where it is used
- Namespace dispatch messages so it doesn't collide with anything

Here is an example on how this would work:

``` go
package main

import (
    "engo.io/ecs"
    "engo.io/engo"
    "fmt"
    "image/color"
)

// ---------------------------------
// Vars

type MainScene struct{}

// ---------------------------------
// Struct functions

// Setup the scene
func (*MainScene) Setup(world *ecs.World) {}

// ---------------------------------
// Private functions

// Prepares project...
func prepare(msg engo.Message) {
    fmt.Println("Hey!")

    // Set the initial scene...
    _ = engo.SetScene(&MainScene{}, engo.SceneSettings{
        Type:         "Main",
        Assets:       []string{"foo.png"},
        Background:   color.White,
    }, true)
}

// Loops project...
func loop(msg engo.Message) {
    fmt.Printf("\nLoop! %v", msg)
}

// ---------------------------------
// Runtime functions

// Init project function
func init() {
    // Set the initial options
    opts := engo.RunOptions{
        Title:  "Foo",
        Width:  1024,
        Height: 640,
        NoRun:  false,
    }

    engo.Init(opts)

    // Listen for events
    engo.Mailbox.Listen("engo.PreparationMessage", prepare)
    engo.Mailbox.Listen("engo.IterationUpdateMessage", loop)
}

// Main project function
func main() {
    engo.Run()
}
```

### DIFF
--- a/engo.go
+++ b/engo.go
@@ -95,10 +95,8 @@ type RunOptions struct {
 	AssetsRoot string
 }
 
-// Run is called to create a window, initialize everything, and start the main loop. Once this function returns,
-// the game window has been closed already. You can supply a lot of options within `RunOptions`, and your starting
-// `Scene` should be defined in `defaultScene`.
-func Run(o RunOptions, defaultScene Scene) {
+// Init initializes the project. You can supply a lot of options within `RunOptions`.
+func Init(o engo.RunOptions) {
 	// Setting defaults
 	if o.FPSLimit == 0 {
 		o.FPSLimit = 60
@@ -118,6 +116,16 @@ func Run(o RunOptions, defaultScene Scene) {
 
 	opts = o
 
+	// TODO: We may need to check this later on for platform
+	InitGl()
+
+	Mailbox = &MessageManager{}
+}
+
+// Run is called to create a window, initialize everything, and start the main loop. Once this function returns,
+// the game window has been closed already. You can supply a lot of options within `RunOptions`, and your starting
+// `Scene` should be defined in `defaultScene`.
+func Run() {
 	// Create input
 	Input = NewInputManager()
 	if opts.StandardInputs {
@@ -138,14 +146,14 @@ func Run(o RunOptions, defaultScene Scene) {
 	// And run the game
 	if opts.HeadlessMode {
 		if !opts.NoRun {
-			runHeadless(defaultScene)
+			runHeadless()
 		}
 	} else {
 		CreateWindow(opts.Title, opts.Width, opts.Height, opts.Fullscreen, opts.MSAA)
 		defer DestroyWindow()
 
 		if !opts.NoRun {
-			runLoop(defaultScene, false)
+			runLoop(false)
 		}
 	}
 }
@@ -203,6 +211,7 @@ func GameHeight() float32 {
 	return gameHeight
 }
 
+// Closes the window
 func closeEvent() {
 	for _, scenes := range scenes {
 		if exiter, ok := scenes.scene.(Exiter); ok {
@@ -217,6 +226,7 @@ func closeEvent() {
 	}
 }
 
-func runHeadless(defaultScene Scene) {
-	runLoop(defaultScene, true)
+// Run the project headless
+func runHeadless() {
+	runLoop(true)
 }
--- a/engo_glfw.go
+++ b/engo_glfw.go
@@ -33,6 +33,35 @@ var (
 	canvasWidth, canvasHeight float32
 )
 
+// WindowResizeMessage is a message that's being dispatched whenever the game window is being resized by the gamer
+type WindowResizeMessage struct {
+	OldWidth, OldHeight int
+	NewWidth, NewHeight int
+}
+
+// IterationUpdateMessage will be used with mailbox
+type IterationUpdateMessage struct {
+	Delta float32
+}
+
+// PreparationMessage will be used with mailbox
+type PreparationMessage struct{}
+
+// Type of message that will be sent
+func (WindowResizeMessage) Type() string {
+	return "engo.WindowResizeMessage"
+}
+
+// Type of message that will be sent
+func (IterationUpdateMessage) Type() string {
+	return "engo.IterationUpdateMessage"
+}
+
+// Type of message that will be sent
+func (PreparationMessage) Type() string {
+	return "engo.PreparationMessage"
+}
+
 // fatalErr calls log.Fatal with the given error if it is non-nil.
 func fatalErr(err error) {
 	if err != nil {
@@ -40,6 +69,7 @@ func fatalErr(err error) {
 	}
 }
 
+// CreateWindow creates a Gl window
 func CreateWindow(title string, width, height int, fullscreen bool, msaa int) {
 	err := glfw.Init()
 	fatalErr(err)
@@ -181,10 +211,12 @@ func CreateWindow(title string, width, height int, fullscreen bool, msaa int) {
 	})
 }
 
+// DestroyWindow destroys the gl window
 func DestroyWindow() {
 	glfw.Terminate()
 }
 
+// SetTitle sets the title of the window
 func SetTitle(title string) {
 	if opts.HeadlessMode {
 		log.Println("Title set to:", title)
@@ -203,8 +235,10 @@ func RunIteration() {
 		glfw.PollEvents()
 	}
 
-	// Then update the world and all Systems
-	currentWorld.Update(Time.Delta())
+	// Announce that the loop happened
+	Mailbox.Dispatch(IterationUpdateMessage{
+		Delta: engo.Time.Delta(),
+	})
 
 	// Lastly, forget keypresses and swap buffers
 	if !opts.HeadlessMode {
@@ -218,17 +252,17 @@ func RunIteration() {
 }
 
 // RunPreparation is called automatically when calling Open. It should only be called once.
-func RunPreparation(defaultScene Scene) {
+func RunPreparation() {
 	Time = NewClock()
 
 	// Default WorldBounds values
 	//WorldBounds.Max = Point{GameWidth(), GameHeight()}
 	// TODO: move this to appropriate location
 
-	SetScene(defaultScene, false)
+	Mailbox.Dispatch(PreparationMessage{})
 }
 
-func runLoop(defaultScene Scene, headless bool) {
+func runLoop(headless bool) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 	signal.Notify(c, syscall.SIGTERM)
@@ -237,7 +271,7 @@ func runLoop(defaultScene Scene, headless bool) {
 		closeEvent()
 	}()
 
-	RunPreparation(defaultScene)
+	RunPreparation()
 	ticker := time.NewTicker(time.Duration(int(time.Second) / opts.FPSLimit))
 
 	// Start tick, minimize the delta
@@ -308,6 +342,7 @@ func SetCursor(c Cursor) {
 	window.SetCursor(cur)
 }
 
+// SetVSync sets Gl vsync
 func SetVSync(enabled bool) {
 	opts.VSync = enabled
 	if opts.VSync {
@@ -317,7 +352,8 @@ func SetVSync(enabled bool) {
 	}
 }
 
-func init() {
+// Initializes application
+func InitGl() {
 	runtime.LockOSThread()
 
 	Grave = Key(glfw.KeyGraveAccent)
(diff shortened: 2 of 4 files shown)


message.go
```diff
@@ -30,11 +30,7 @@ func (mm *MessageManager) Listen(messageType string, handler MessageHandler) {
 	mm.listeners[messageType] = append(mm.listeners[messageType], handler)
 }
 
-// WindowResizeMessage is a message that's being dispatched whenever the game window is being resized by the gamer
-type WindowResizeMessage struct {
-	OldWidth, OldHeight int
-	NewWidth, NewHeight int
+// Remove removes all listeners of a message type
+func (mm *MessageManager) Remove(messageType string) {
```
otraore: WHy wouldn't just allow for removing one listener and also functionality to remove all. This name is confusing.

engo.go
```diff
@@ -118,6 +116,16 @@ func Run(o RunOptions, defaultScene Scene) {
 
 	opts = o
 
+	// TODO: We may need to check this later on for platform
+	InitGl()
```
otraore: This is totally unnecessary, there's a reason why go has an [init function](https://golang.org/doc/effective_go.html#init).

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
 	canvasWidth, canvasHeight float32
 )
 
+// WindowResizeMessage is a message that's being dispatched whenever the game window is being resized by the gamer
+type WindowResizeMessage struct {
```
otraore: glfw isn't the only platform we support so this being in this file breaks all the other platforms.

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+	OldWidth, OldHeight int
+	NewWidth, NewHeight int
+}
+
+// IterationUpdateMessage will be used with mailbox
+type IterationUpdateMessage struct {
```
otraore: Same as above

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+	Delta float32
+}
+
+// PreparationMessage will be used with mailbox
+type PreparationMessage struct{}
```
otraore: Same as above

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+
+// Type of message that will be sent
+func (WindowResizeMessage) Type() string {
```
otraore: Same as above

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+	return "engo.WindowResizeMessage"
+}
+
+// Type of message that will be sent
+func (IterationUpdateMessage) Type() string {
```
otraore: Same as above

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+	return "engo.IterationUpdateMessage"
+}
+
+// Type of message that will be sent
+func (PreparationMessage) Type() string {
```
otraore: Same as above

engo_glfw.go
```diff
@@ -33,13 +33,43 @@ var (
...
+
 // fatalErr calls log.Fatal with the given error if it is non-nil.
 func fatalErr(err error) {
 	if err != nil {
 		log.Fatal(err)
 	}
 }
 
+// CreateWindow creates a Gl window
 func CreateWindow(title string, width, height int, fullscreen bool, msaa int) {
```
otraore: same as above

engo_glfw.go
```diff
@@ -317,7 +352,8 @@ func SetVSync(enabled bool) {
 	}
 }
 
-func init() {
```
otraore: Completely unnecessary

message.go
```diff
@@ -30,11 +30,7 @@ func (mm *MessageManager) Listen(messageType string, handler MessageHandler) {
 	mm.listeners[messageType] = append(mm.listeners[messageType], handler)
 }
 
-// WindowResizeMessage is a message that's being dispatched whenever the game window is being resized by the gamer
-type WindowResizeMessage struct {
```
otraore: Like I said before, this breaks other platforms.

message.go
```diff
@@ -30,11 +30,7 @@ func (mm *MessageManager) Listen(messageType string, handler MessageHandler) {
...
-	OldWidth, OldHeight int
-	NewWidth, NewHeight int
+// Remove removes all listeners of a message type
+func (mm *MessageManager) Remove(messageType string) {
+	delete(mm.listeners, messageType)
 }
-
-// Type returns the type of the current object "WindowResizeMessage"
-func (WindowResizeMessage) Type() string { return "WindowResizeMessage" }
```
otraore: Like I stated before, this breaks other platforms

scene.go
```diff
@@ -11,14 +11,14 @@ var scenes = make(map[string]*sceneWrapper)
 // Scene represents a screen ingame.
 // i.e.: main menu, settings, but also the game itself
 type Scene interface {
-	// Preload is called before loading resources
-	Preload()
```
otraore: Bad idea, I'll explain why and amongst other reasons I think this PR is a bad idea.

engo.go
Sendoushi: Yes, I understand but this way you can specify on your project what to do.
Engo is a `lib` you should be able to set the project how you want it and maybe you have some more stuff to do in a `init`.

scene.go
```diff
@@ -11,14 +11,14 @@ var scenes = make(map[string]*sceneWrapper)
...
-
 	// Setup is called before the main loop
 	Setup(*ecs.World)
+}
 
-	// Type returns a unique string representation of the Scene, used to identify it
-	Type() string
+type SceneSettings struct {
```
otraore: Also realized documentation is heavily lacking.

engo_glfw.go
Sendoushi: Oh. Makes sense. I made a search and didn't found anything... I'll remove this part then!

engo.go
otraore: Engo is a game engine not a game library. [Difference between the two](http://gamedev.stackexchange.com/questions/10770/whats-the-difference-between-a-library-and-an-engine)

engo_glfw.go
Sendoushi: I make here the same point. `Engo` is a `lib` and as a developer I want the flexibility to setup my project as I want. This is the issue I was talking about on the other issue. There is no need for the `render` to know about scenes or worlds.

If I want to use another pattern with different kind of approach I can't because I have to set this up with ecs elements.

message.go
Sendoushi: Yes. I'm on it...

message.go
Sendoushi: Yes that would be the next step. Maybe have instead a `RemoveType` and a `Remove`

engo.go
Sendoushi: Ok then. I understand the difference. Thought of it as a lib.

engo_glfw.go
otraore: See my response above, and I don't see how you want `flexibility` yet still removed the `Preload` method from the Scene interface.

Sendoushi: Because it isn't needed. You already know the assets so why don't you just load it in the setup? But then maybe you have a point and maybe `Preload` makes sense.
These are just a set of propositions.

otraore:
I overall personally this PR is a bad idea. For one it removes `Preload` method on the `Scene` interface. This alone takes a lot of flexibility away from the user. I disagree with removing it. There's no need for a `loop` message, each system gets an update message. And I see no reasons for separating `engo.Run()`  into two functions being `engo.Run` and engo.Init()`. If your ultimate goal is to remove the need to use ECS in engo, I think you're going about it the wrong way.  I also think you dislike ECS because you don't understand it, which is an issue. We're discussing how we can teach new users on how to get used to ECS.

message.go
Sendoushi: On it...

engo_glfw.go
Sendoushi: On it...

engo_glfw.go
Sendoushi: On it...

engo_glfw.go
Sendoushi: On it...

engo_glfw.go
Sendoushi: On it...

engo_glfw.go
```diff
@@ -317,7 +323,8 @@ func SetVSync(enabled bool) {
...
+// Initializes application
+func InitGl() {
 	runtime.LockOSThread()
 
 	Grave = Key(glfw.KeyGraveAccent)
```
EtienneBruines: These should be constant, and not within this method. This is something related to the `input_data.go` file - not PR related, but should change.

engo_glfw.go
```diff
@@ -317,7 +323,8 @@ func SetVSync(enabled bool) {
 	}
 }
 
-func init() {
+// Initializes application
+func InitGl() {
 	runtime.LockOSThread()
```
EtienneBruines: This method should not be here, but when opening the OpenGL window. And maybe even related to OpenGL calls? Not sure about that. Having it here does not make sense directly, IMHO. 

Also, not PR related, but if this one is moved as well, this method is empty and can be removed entirely. No need to rename or discuss about `init` vs `InitGl`.

EtienneBruines: `InitGl` would be called `InitGL` though, because `GL` is an acronym. Like `HTML`.

EtienneBruines:
I have not looked at the code at this point, yet. Let me first comment on every "bulletpoint" you mentioned - as far as I can. 

* Remove function in Mailbox (untested)
 * ?

* Common Mailbox for all (I think that having a global Mailbox will be easier to use and it helps with background stuff for example)
  * We do have a common mailbox? The difference is that a "scene"-change makes sure you don't waste code on listening for messages that don't appear. It might be true though, that because the "events" don't get fired, listening does not form any performance bottleneck. Except maybe for things being in-memory when they don't have to be any more. 

* SceneSettings instead of interface functions
  * Neutral opinion. 

* Removed mailbox from scene. Instead, it may use the global one.
  * Same as above. 

* InitGl instead of using the global func init()
  * The method may even be removed entirely; not sure about that. Worth investigating. See also comment in code. 

* Separate Init from Run
  The idea is that `Run` does everything you'd want it to. Not sure *why* you'd want to split it?

* Dispatch preparation and iteration messages (Render shouldn't know about scenes, worlds... It will help unpatternize the engine)
  * I think this is the part where @otraore has his objections. I think it won't help "unpatternize" the engine - I think it will just follow an *other* pattern. 

* Move WindowResizeMessage to the place where it is used
  * Yeah, you didn't do that ;) It's still where it has always been. 

* Namespace dispatch messages so it doesn't collide with anything
  * Neutral about this; don't really care what they're called ;-). 

The thing I *dislike* personally about your code-snippet, is the fact that `engo.Run` now becomes state-dependant. I like the fact that (in the current version) you simply give parameters whatever you like, instead of "registering" methods.

scene.go
paked: I think it might be good for @Sendoushi to explain his thinking for this one.

otraore: We already talked on gitter, he ended up changing his mind too.

Sendoushi: Yes. I changed my mind. I was trying to simplify but running against myself when I talk about flexibility. This method gives flexibility to the developer.

engo_glfw.go
Sendoushi: Yes but that is related to the other PR. I totally agree. I used the actual code.

otraore:
I've had second thoughts about letting the user have a custom method for every fps iteration. I'm not sure about it being in the mailbox but something like
 ```go
engo.OnUpdate(func() {
   fmt.Println("Frame") 
})
``` 

"OnUpdate" is up for change just an example. But a Mailbox message wouldn't bother me either.

Noofbiz:
#513 Separates out the requirement to use common and ecs in order to use the other aspects of engo.

PR #367, opened 2016-06-16
Hotfix retina display

EtienneBruines:
Deprecates / replaces https://github.com/EngoEngine/engo/pull/352. 

This is the "correct" fix for the retina displays.

### DIFF
--- a/engo_glfw.go
+++ b/engo_glfw.go
@@ -269,14 +269,6 @@ Outer:
 	ticker.Stop()
 }
 
-func GameWidth() float32 {
-	return gameWidth
-}
-
-func GameHeight() float32 {
-	return gameHeight
-}
-
 func CursorPos() (x, y float32) {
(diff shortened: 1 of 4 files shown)


paked:
LGTM once fixed. Pasting error below.

```
./engo_glfw.go:272: GameWidth redeclared in this block
	previous declaration at ./engo.go:197
./engo_glfw.go:276: GameHeight redeclared in this block
	previous declaration at ./engo.go:202
```

engo_glfw.go
```diff
@@ -262,8 +269,17 @@ Outer:
 	ticker.Stop()
 }
 
-func CursorPos() (x, y float64) {
-	return window.GetCursorPos()
+func GameWidth() float32 {
```
laher: This is also defined in engo.go. Does not compile.

Issue #306, opened 2016-05-16
Try out iOS


EtienneBruines:
Someone with an iPhone and XCode should try to run it. 

I've read/heard that `ios` is _not_ the build-tag for iOS. Instead you use `amd64 darwin` (note the space in between). If we were to use that, maybe we could get it to build. 

Up for grabs.

otraore:
Grabbing later in the day

EtienneBruines:
Be sure to share your findings at gitter :smile:

otraore:
Update we got it to build. Sadly I wasn't able to test it out on the emulator. We think it's because of https://github.com/golang/go/wiki/Mobile#ios-simulator . This might differ on a iOS device.

paked:
Do we need to find someone with an iOS device in order to continue testing?

otraore:
Believe so, or I can set up the simulator for it according to x/mobile's suggestions

paked:
@otraore How did this go?

PR #370, opened 2016-06-22
Added build support for iOS

otraore:
Along with https://github.com/EngoEngine/gl/pull/11 , This pull request adds in build support for iOS. Due to current limitations with gomobile and the iOS emulator, I'm not able to confirm wether the built applications work as intended. In this PR I've purposely left a built application within the demos/pong/ folder called pong.app. Anyone is welcomed to try to run it on their iOS devices if they have one. This attempts to fix: #306 once merged. I'd also like to make a side note stating that I've tried and gotten engo to build on iOS before but at the cost of breaking OSX support because of faulty build tags. So I'd like to thank @hajimehoshi and his recent efforts to make ebitten build on iOS which included correct and working build tags.


engo_mobile.go
```diff
@@ -1,4 +1,4 @@
-//+build android
+//+build android darwin,arm darwin,arm64
```
hajimehoshi: Maybe you would need `ios` tag here

engo_glfw.go
```diff
@@ -1,4 +1,4 @@
-//+build !netgo,!android
+// +build darwin,!arm,!arm64 linux windows
```
hajimehoshi: Maybe you would need a line `// +build !ios` for iOS emulators (where build tags are `darwin` & `amd64` & `ios`).
Additionally, you need `// +build !android`.

paked:
Looks like you've accidentally committed `pong.app` into the repository, would you mind removing it?

otraore:
@paked if you read the OP I mentioned it. 
>In this PR I've purposely left a built application within the demos/pong/ folder called pong.app. Anyone is welcomed to try to run it on their iOS devices if they have one.

engo_glfw.go
otraore: I'll take a look into it. I'm not sure you'd need a build tag for the emulator. Also why would I need to exclude android. In this line It's including rather than excluding so I'm not sure I'd need to include `!android`. Thanks for the comment though, appreciate it.

hajimehoshi: > I'll take a look into it. I'm not sure you'd need a build tag for the emulator. 
`gomobile` makes binaries for emulators (amd64) whether you want or you don't. (Note that it looks like specifying architecture on `-target` like `-target ios:arm` of `gomobile` doesn't work for iOS. This seems TBD).

> Also why would I need to exclude android.
It's because android is one of linux machines and if you specify `linux` for build tags, this must include android machine unless you don't exclude android.

Luukdegram:
> I've purposely left a built application within the demos/pong/ folder called pong.app

If I remember correctly, you'd need a provisioning profile included with the app to install it on an authorized device.

EtienneBruines:
Building fails on travis.

paked:
This is the error:
```
common/audio_player.go:86: undefined: ReadSeekCloser
```

Issue #294, opened 2016-05-13
Check for unnecessary converions during CI


mewmew:
We should consider using `unconvert` to check for unnecessary conversions during CI. This would further help improve the code quality of engo as time goes by.

https://github.com/mdempsky/unconvert

Example PR: https://github.com/EngoEngine/engo/pull/293

PR #373, opened 2016-06-25
Check for unnecessary conversions during tests

otraore:
Fixes: https://github.com/EngoEngine/engo/issues/294


EtienneBruines:
LGTM.

PR #382, opened 2016-08-16
MouseSystem now skips hidden entities

EtienneBruines:
Currently, even though you hide something, it can receive hover/click events. 

This is unwanted, because you can click something that's explicitely hidden. If you want something invisible, but clickable, you should set alpha to 0, instead of defining `Hidden`, IMHO. 

@paked @otraore ?


otraore:
I agree, LGTM. I'll merge when builds pass.

PR #390, opened 2016-11-15
refactor adventure demo to use a SpeedSystem

yarbelk:
Idea taken from the pong demo - precurser to adding collisions for the trees, I wanted the speed and movement separated from  the controls.


EtienneBruines:
Looks great to me! Thank you for your contribution and apologies for the delay.

PR #393, opened 2016-11-21
common: animation improvements

fjl:
Please see commit messages for change details.

### DIFF
--- a/common/animation.go
+++ b/common/animation.go
@@ -20,7 +20,7 @@ type AnimationComponent struct {
 	Rate             float32               // How often frames should increment, in seconds.
 	index            int                   // What frame in the is being used
 	change           float32               // The time since the last incrementation
-	def              string                // The default animation to play when nothing else is playing
+	def              *Animation            // The default animation to play when nothing else is playing
 }
 
 func NewAnimationComponent(drawables []Drawable, rate float32) AnimationComponent {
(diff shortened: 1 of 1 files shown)


fjl:
Is there a reason why `NextFrame` and `Cell` are exported? They're only used by AnimationSystem and will crash if used incorrectly.

common/animation.go
```diff
@@ -80,41 +94,39 @@ func (ac *AnimationComponent) NextFrame() {
...
-	delete := -1
-	for index, e := range a.entities {
-		if e.BasicEntity.ID() == basic.ID() {
-			delete = index
-			break
-		}
-	}
-	if delete >= 0 {
-		a.entities = append(a.entities[:delete], a.entities[delete+1:]...)
+	if a.entities != nil {
```
EtienneBruines: If you add this, then you won't have to do these checks. 

```go
func (a *AnimationSystem) New(*ecs.World) {
    a.entities = make(map[ecs.BasicEntity]animationEntity)
}
```

EtienneBruines:
Very nice, and thank you for your contribution!

The `master` branch changed a bit (just merged a PR), so you may want to update your local branch before doing any changes. 

I have reviewed your PR and it looks pretty nice. There was one thing I commented, maybe you want to have a look at that.

common/animation.go
fjl: That's true, but it also means I can't add entities before adding the system to World. It'd be nice to keep that working.

fjl:
@EtienneBruines can you answer my question about Cell and NextFrame

otraore:
@fjl In response to your question, I don't believe `Cell()` will crash the program if used incorrectly. It's just a reference to the current animation frame which can be useful . As for `NextFrame(`) it also just allows the user more flexibility. For example, they can stop the animation loop and manually trigger the frames. If you still think they either of them should be unexported, you can create an issue where we can discuss it.

With that said, On behalf of the Engo team I'd like to thank you for your contribution. 
LGTM.

PR #399, opened 2017-01-01
Isometric Support

thecarp:
This branch adds isometric and staggered isometric map support, and the ability to define new mappings.

Mapping points to map coordinates is abstracted into two complementary functions MapToPosition and PositionToMap  functions, which can be defined according to the map orientation. A setupOrientation() function provides default mappings for orthogonal, isometric, and staggered. Hex mapping should just be a matter of defining its position functions. 

Both adventure and tilemap demos work without change using their existing assets.

### DIFF
--- a/common/level.go
+++ b/common/level.go
@@ -93,8 +93,6 @@ func (lvl *Level)setupOrientation() error {
 			return
 		}
 		lvl.PointToMap = func(p engo.Point) (m *engo.Point) {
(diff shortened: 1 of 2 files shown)


thecarp:
Closing this request for now. Using this branch a bit more I found that it needed some work still.
Will submit an issue and new request in a few days when I have worked out a bit more and am happy with a demo.

otraore:
Looking forward to it @thecarp .

PR #404, opened 2017-01-21
All Common Components now have GetCommonComponent methods

coderconvoy:
Each CommonComponent now has a GetCommonComponent method.

eg 
```
func (sc *SpaceComponent) GetSpaceComponent() *SpaceComponent{
    return sc
}
```

All common systems, with Add methods now also have an AddByInterface method, using the GetCommonComponent methods as required methods.  These are of course propogated to the container of the objects, so each entity, with a component automaticalle implements the required interface.

I have also created a Pull request in ecs to make sure GetBasicEntity is also available

At the user end, they can create the objects in the same way as before, but now they can add the object to the system using the AddByInteface()

eg
```
sys.AdByInterface(ob)
```

which is much simpler than 
```
sys.Add(ob.BasicEntity,ob.SpaceComponent,on.OtherComponent)
```

And doesn't require the user to remember the order.

### DIFF
--- a/common/animation.go
+++ b/common/animation.go
@@ -94,6 +94,11 @@ func (ac *AnimationComponent) NextFrame() {
 	}
 }
 
+// GetAnimationComponent provides container classes with interface for access to Component
+func (ac *AnimationComponent) GetAnimationComponent() *AnimationComponent {
+	return ac
+}
+
 // AnimationSystem tracks AnimationComponents, advancing their current animation.
 type AnimationSystem struct {
 	entities map[ecs.BasicEntity]animationEntity
@@ -110,6 +115,16 @@ func (a *AnimationSystem) Add(basic *ecs.BasicEntity, anim *AnimationComponent, 
 		a.entities = make(map[ecs.BasicEntity]animationEntity)
 	}
 	a.entities[*basic] = animationEntity{anim, render}
+}
+
+// AddByInterface is a Simpler outward facing equivilant of Add, as long as the object meets the interface it can be added safely. Each Component, should provide these methods, so that entities automatically fulfull the interface
+func (a *AnimationSystem) AddByInterface(ob interface {
+	GetBasicEntity() *ecs.BasicEntity
+	GetAnimationComponent() *AnimationComponent
+	GetRenderComponent() *RenderComponent
+}) {
+	a.Add(ob.GetBasicEntity(), ob.GetAnimationComponent(), ob.GetRenderComponent())
+
 }
 
 // Remove stops tracking the given entity.
--- a/common/audio.go
+++ b/common/audio.go
@@ -37,10 +37,21 @@ func (ac *AudioComponent) SetVolume(volume float64) {
 	ac.player.SetVolume(volume * MasterVolume)
 }
 
+func (ac *AudioComponent) GetAudioComponent() *AudioComponent {
+	return ac
+}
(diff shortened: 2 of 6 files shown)


coderconvoy:
I'm not sure what this comment means, or how to respond to it. I quite new to git.

common/animation.go
```diff
@@ -112,6 +117,16 @@ func (a *AnimationSystem) Add(basic *ecs.BasicEntity, anim *AnimationComponent,
 	a.entities[*basic] = animationEntity{anim, render}
 }
 
+// AddByInterface is a Simpler outward facing equivilant of Add, as long as the object meets the interface it can be added safely. Each Component, should provide these methods, so that entities automatically fulfull the interface
+func (a *AnimationSystem) AddByInterface(ob interface {
```
otraore: Declare the interface elsewhere so it's not inlined

common/audio.go
```diff
@@ -37,12 +37,23 @@ func (ac *AudioComponent) SetVolume(volume float64) {
 	ac.player.SetVolume(volume * MasterVolume)
 }
 
+func (ac *AudioComponent) GetAudioComponent() *AudioComponent {
```
otraore: Documentation like the other functions

common/audio.go
```diff
@@ -37,12 +37,23 @@ func (ac *AudioComponent) SetVolume(volume float64) {
...
+	return ac
+}
+
 type audioEntity struct {
 	*ecs.BasicEntity
 	*AudioComponent
 	*SpaceComponent
 }
 
+//interface for adding audio entities using AddByInterface
```
otraore: Not [godoc style documentation](https://blog.golang.org/godoc-documenting-go-code)

common/collision.go
```diff
@@ -126,6 +131,10 @@ type CollisionComponent struct {
 	Collides    bool // Collides is true if the component is colliding with something during this pass
 }
 
+func (cc *CollisionComponent) GetCollisionComponent() *CollisionComponent {
```
otraore: This is also missing documentation

common/collision.go
```diff
@@ -126,6 +131,10 @@ type CollisionComponent struct {
...
+	return cc
+}
+
 type CollisionMessage struct {
```
otraore: Documentation

common/collision.go
```diff
@@ -139,6 +148,12 @@ type collisionEntity struct {
 	*SpaceComponent
 }
 
+type Collisioner interface {
```
otraore: You switched naming conventions, this can be confusing. You went from `SystemInterface` to `Systermer`

coderconvoy: These are good thoughts.  I did change styles half way through as I was doing it, partly to see which I prefered.  I've been using it for a month now, and I think I know how it can be done better.

I want to get rid of the "Get"  part  of Get__Component etc , but the problem is, the parent object will hold the object and inherit the method, so I can't call it the same name as the object as this will cause confusion possibly to the compiler and definitly to some programmers.

I'm going to re-edit and make a new pull request shortly

coderconvoy:
I've put up the new pull request now. so ignore this one.

coderconvoy:
OK, found, I had accidently edited a binary file in the second attempts, so I have closed that one too.

#414 should be the one you want now

PR #410, opened 2017-02-24
New method SpaceComponent.GetCenter()

countcb:
This method gets the center position of the space component instead of its top-left point.
The go idiomatic way would be to rename Center(p) to SetCenter(p) and GetCenter() to Center(). But since that change would break backwards-compability I went for GetCenter() as the new method name.

If you would like the methods to be renamed to SetCenter(p) and Center() I can change the commit accordingly. As a first time contributor I think it is not up to me to decide whether or not a change like this should break backwards-compability. :)


countcb:
Will repost this with own branch instead of using master branch.

PR #411, opened 2017-02-24
Fixed legacyShader.Draw() to draw Circle borders in the correct color.

countcb:


countcb:
Rebased this onto the current master branch.

PR #412, opened 2017-02-24
Added Get__Component methods, and interfaces, and all the AddByInterf…

coderconvoy:
…ace methods

This is the updated pull request, I hope I have done this right, I'm relatively new to using Git.

This Pull request adds a File full of useful interfaces.  called interfaces.go, and adds the Get__Component Methods to each Component Type.
The methods are still called Get__ as I removeing the Get, made a name collision, within the parent objects.  

The "AddByInterface" methods which allow Systems to add an Entity as one parameter long as it has the correct components are all directly after the Add method which they call.

interface_test.go has also been added, and this tries all of the AddByInterface methods and checks the objects are added, and removable.

For this update to be effective, it will also be helpful to pull the request on Engo.ecs to give ecs.BasicEntity a GetBasicEntity Method, then all of this will work together.

### DIFF
--- a/common/animation.go
+++ b/common/animation.go
@@ -112,6 +112,11 @@ func (a *AnimationSystem) Add(basic *ecs.BasicEntity, anim *AnimationComponent, 
 	a.entities[*basic] = animationEntity{anim, render}
 }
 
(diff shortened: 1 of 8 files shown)


coderconvoy:
Ooops, I accidently changed a binary file, and some stuff in Adventure.go as I was experimentting with that a few days ago, I didn't realise that would get pulled into the commit.

PR #413, opened 2017-02-24
New method SpaceComponent.Center()

countcb:
This method gets the center position of the space component instead of its top-left point.
The go idiomatic way would be to rename Center(p) to SetCenter(p) and GetCenter() to Center(). But since that change would break backwards-compability I went for GetCenter() as the new method name.

If you would like the methods to be renamed to SetCenter(p) and Center() I can change the commit accordingly. As a first time contributor I think it is not up to me to decide whether or not a change like this should break backwards-compability. :)

### DIFF
--- a/common/collision.go
+++ b/common/collision.go
@@ -25,6 +25,15 @@ func (sc *SpaceComponent) Center(p engo.Point) {
 	sc.Position.Y = p.Y - yDelta
 }
 
+// GetCenter gets the center position of the space component instead of its
+// top-left point (this avoids doing the same math each time in your systems)
+func (sc *SpaceComponent) GetCenter() engo.Point {
+	xDelta := sc.Width / 2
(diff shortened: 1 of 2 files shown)


otraore:
I would prefer you go to idiomatic route since we're pre 1.0 breaking the API for changes like this is fine. I appreciate that you asked first though.

countcb:
Ok. Made the changes and updated this PR. :)

otraore:
Thank you for your contributions.

countcb:
I don't know why but the changes I made regarding the go idiomatic way of naming the methods got lost...
I'll make a PR which renames the methods. Sorry for the confusion...

otraore:
Woah, that's weird. I was sure you updated them. If you can submit a new PR that would be helpful.

PR #414, opened 2017-02-24
Get__Component and AddByInterface

coderconvoy:
Attempt 3:

Created a new file called interfaces.go, which defines interfaces to requiring Get__Component methods to allow AddByInterface methods to recieve an interface, instead of a list of components

Added AddByInterface methods to all common systems that have an Add method.  These receive an interface defined in interfaces.go, to make adding an entity to a system work with a single parameter.

Added Get__Component method to each component so that any containing interface automatically satisfies interfaces required by AddByInterface methods.

created a new file called interfaces_test.go, to make sure all new code is tested.  Could not add fail tests easily as the whole thing is typesafe, so would not compile.

### DIFF
--- a/common/animation.go
+++ b/common/animation.go
@@ -112,6 +112,11 @@ func (a *AnimationSystem) Add(basic *ecs.BasicEntity, anim *AnimationComponent, 
 	a.entities[*basic] = animationEntity{anim, render}
 }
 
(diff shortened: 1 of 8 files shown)


countcb:
This is the approach I used in my own systems. 
Looking forward to this getting merged :) 
Thanks for the idea and work.

otraore:
This depends on https://github.com/EngoEngine/ecs/pull/28#pullrequestreview-24387686.  I can't merge it in yet with using a temporary `SafeBasic`. Once the ECS PR is merged and the tests are passing again, I'll merge it in.

Issue #392, opened 2016-11-20
Impossible to scroll on subsequent scenes


johnny-morrice:
It's impossible to do CameraSystem based scrolling on all but the first scene.  (I have tested with EdgeScroller but I believe this applies across the board)

Why?

On shader initialization, the global shader instances (e.g. DefaultShader) grab a reference to the first CameraSystem they find (See *DefaultShader Setup method).

When the user creates a new RenderSystem for use in a second scene ([as per the examples](https://github.com/EngoEngine/engo/blob/master/demos/scenes/scenes.go)), a new CameraSystem is created, but the references in the shaders are not updated.

Here is a quick and stupid fix whereby a single CameraSystem instance persists across all RenderSystem instances:

```
func (rs *RenderSystem) New(w *ecs.World) {
	rs.world = w

	if __camera == nil {
		__camera = &CameraSystem{}
	}

	w.AddSystem(__camera)

	if !engo.Headless() {
		initShaders(w)
		engo.Gl.Enable(engo.Gl.MULTISAMPLE)
	}

	engo.Mailbox.Listen("renderChangeMessage", func(engo.Message) {
		rs.sortingNeeded = true
	})
}

var __camera *CameraSystem
```

This is stupid but you'll find that one can now scroll in scenes after the first.

Another suggestion is that one could fix the bug by ensuring the CameraSystem instances are not shared across global variables such as DefaultShader.  Perhaps by turning those global variables into instance variables or methods.  As I'm not a core contributor I don't know what design you have in mind.

countcb:
Added a PR with an idea to fix this here: https://github.com/EngoEngine/engo/pull/420

PR #420, opened 2017-03-01
Proposal: Potential fix for Issue #392

countcb:
New method addCamerSystemOnce to ensure, that only one CameraSystem is added to the world regardless of how many RenderSystems are created.

Protect RenderSystem and Shader initialzation with mutex

Open for discussion. If this is not the direction you want to take to fix this that's ok.
Maybe it would be good to add this non the less because it fixes the problem and can be removed/replaced without affecting the API.


countcb:
Resolved the merge conflict with the master branch (caused by the new math package import).

otraore:
LGTM, Will merge once tests pass.

otraore:
Thank you for your contribution

PR #422, opened 2017-03-04
Tmx encodings

thecarp:
Updated tmx loading to support all encodings and compressions in the tiled spec.

Refactored the decoder out to a new function and added a unit test for it.

Tested against adventure and tilemap demos.


otraore:
After the changes requested are met, LGTM. 

I look forward to merging this in. Thank you for your contribution :D.

otraore:
Thank you again for your contribution :D

PR #423, opened 2017-03-04
Fixed Mouse Action being overwritten

AnkithD:
On some systems a right click also triggers a CursorPos callback, and left click with rapid mouse movement also triggers a CursorPos Callback. Since Release and Press actions are usually prefered to be reported over the Move action, an if statement ensures that the mouse button clicks are reported over mouse movement. Edit: This was written in order to solve Issue#417


AnkithD:
Im not sure why Travis CI Build test failed...

coderconvoy:
The problem is travis CI, not your code. There was a problem with the update to go 1.8

PR #424, opened 2017-03-04
Collision system now uses CollisionGroups +test

coderconvoy:
**Not backwards compatible**

The basic aim of this Pull request is to allow multiple  different kinds of Main-collisionEntities to collide differently with different kinds of non-Main collisionEntities.

To do this, I have replaced, Solid, and Main (bool), in CollisionComponent with  Main and Group CollisionGroup
``` 
type CollisionGroup byte
```

These bytes will be set by the user to allow a bitwise "and" between e1.Main, & e2.Group, and if the result is non zero the items are considered to collide.  Using byte allows up to 8 different collision groups, (I guess we can make it more with an int if prefered. I only needed 3 in my project so far)

To test if the collision should be solid, the CollisionSystem stores another byte called Solids. which will then be "and"ed with the result of the previous result ie (sys.Solids & e1.Main & e2.Group)  If the result is still non zero then the collision is Solid, and 'e1' will move to accomodate.

I expect the user to create a bitmask const set like this 

```
const (
   Ball = 1 << iota
   Bat // 2
   Animal  //4
)
```
and add create Collision components, like this
```
ballcomp := CollisionComponent{
    Main :Ball,
    Group : 0,
}

wallComp :=CollisionComponent{
    Main : 0
    Group : Ball | Bat | Animal
}
```
And to set the system solids at creation
```
collisionSys := &CollisionSystem{
    Solids:Bat |Ball,
}
```

Lastly,  ob.Collides is now also a collision group, so that if anyone wants to check it, they can now read for specific types of collision.
```
if ob.Collides & Ball != 0 {
//do ball collision
}
if ob.Collides & Bat != 0{
//do bat collision
}
```


otraore:
You're free to resolve the merge conflicts so it's up to date with master again. As for my thoughts, I definitely agree with the issue that the pull request is trying to solve. I think this solution works but since it is a breaking change I thought I'd see what either @paked or @EtienneBruines has to say about it.

coderconvoy:
Makes sense to me.

Noofbiz:
I really like this idea. @paked or @EtienneBruines what do you guys think about getting it in?

paked:
I like the idea of splitting things into collision groups. Seems like a much nicer model than 'main' or 'solid' ;)

PR #426, opened 2017-03-07
Gopherjs compatible Sin, Cos, Sqrt

countcb:
Added wrapper package engo/math.
    
This package currently is a wrapper to github.com/engoengine/math.
It makes it possible to exchange the wrapped math package completely or only change single methods while maintaining a stable engo api.
    
This package should be used by engo itself and can be used by it's users.

Added JavaScript compatible versions of: Cos, Sin and Sqrt

If you don't like the wrapper approach the changes would have to be made in github.com/engoengine/math.
After reading the readme I wasn't sure how much this package can be changed.
The wrapper approach has the advatage, that engo becomes independent of the used math package and can maintain a stable API.

If this is not the way this should be solved it is ok for me though :)


otraore:
Thanks for the PR but please change the `engo.io/math` to `github.com/engoengine.com/math`. The reason is that projects under `engo.io/*` are made by EngoEngine members. `EngoEngone/math` is a fork that originated from a repository that no longer existed but engo currently depends on it. So for the time being, it will use the github.com import path.

countcb:
I don't understand? The import path is still "github.com/engoengine/math".
"engo.io/math" is the new package I wrote that acts as a wrapper to "github.com/engoengine/math".
So the only files that are importing "github.com/engoengine/math" are currently under "engo.io/math"

This makes the engo api independent of the external package.

> EngoEngone/math is a fork that originated from a repository that no longer existed but engo currently depends on it. 

I know. That's exactly the reason I made this wrapper :)

Do you not want to go the wrapper route? From your comment I'm not sure you noticed that it is just a wrapper?

otraore:
That was misinterpretation on my part, I'm fine with proceeding with this route.

otraore:
Sorry for the late follow-up but thank you for your contribution. We really appreciate it.

PR #439, opened 2017-05-12
Adding EntityExists function to RenderSystem

umarquez:
This function will be used to check if an entity exists (and where) into the entities slice, this will prevents entity's duplicity in the system that can causes a double render process for some entities if these was added twice.

This function could also be called from the Remove function to get the position of the entity that will be remove from the system.

### DIFF
--- a/common/render.go
+++ b/common/render.go
@@ -136,44 +136,58 @@ func addCameraSystemOnce(w *ecs.World) {
 }
 
 func (rs *RenderSystem) Add(basic *ecs.BasicEntity, render *RenderComponent, space *SpaceComponent) {
-	// Setting default shader
-	if render.shader == nil {
-		switch render.Drawable.(type) {
-		case Triangle:
-			render.shader = LegacyShader
-		case Circle:
-			render.shader = LegacyShader
-		case Rectangle:
-			render.shader = LegacyShader
-		case ComplexTriangles:
-			render.shader = LegacyShader
-		case Text:
-			render.shader = TextShader
-		default:
-			render.shader = DefaultShader
+	if rs.EntityExists(basic) == -1 {
+		// Setting default shader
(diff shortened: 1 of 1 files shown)


common/render.go
```diff
@@ -136,44 +136,58 @@ func addCameraSystemOnce(w *ecs.World) {
...
-			render.shader = LegacyShader
-		case Rectangle:
-			render.shader = LegacyShader
-		case ComplexTriangles:
-			render.shader = LegacyShader
-		case Text:
-			render.shader = TextShader
-		default:
-			render.shader = DefaultShader
+	if rs.EntityExists(basic) == -1 {
```
EtienneBruines: [Idiomatic Go](https://talks.golang.org/2013/bestpractices.slide#2) would be to put the exceptional case on the top (with a return), and then the 'regular logic' behind it. 

```go
if rs.EntityExists(basic) >= 0 {
    return
}

// Setting default shader
// ...
```

common/render.go
```diff
@@ -136,44 +136,58 @@ func addCameraSystemOnce(w *ecs.World) {
...
-		case Text:
-			render.shader = TextHUDShader
-		default:
-			render.shader = HUDShader
+// EntityExists looks if the entity is already into the System's entities. It will return the index >= 0 of the object into de rs.entities or -1 if it could not be found.
+func (rs *RenderSystem) EntityExists(basic *ecs.BasicEntity) int {
+	for index, entity := range rs.entities {
+		if entity.ID() == basic.ID() {
+			return index
+			break
```
EtienneBruines: `break` is not needed after `return`, because we're already stopping executing the loop (due to `return`).

umarquez:
@EtienneBruines Fixed!

Noofbiz:
Thank you!

PR #442, opened 2017-06-17
flipped the signs on collision.go initial bounds for AABB()

### LINKED ISSUE #441: SpaceComponent.AABB() doesn't update properly
Hey all, I was trying to generalize some code to only count a mouse as hovering over an image if the alpha being hovered over was zero.

I've gotten it working for a non-rotated image, but I'm starting to dive in to rotated images now, and came across collision.go:44: 

`func (sc SpaceComponent) AABB() engo.AABB {
	if sc.Rotation == 0 {
		return engo.AABB{
			Min: sc.Position,
			Max: engo.Point{sc.Position.X + sc.Width, sc.Position.Y + sc.Height},
		}
	}

	corners := sc.Corners()

	var (
		xMin float32 = -math.MaxFloat32
		xMax float32 = math.MaxFloat32
		yMin float32 = -math.MaxFloat32
		yMax float32 = math.MaxFloat32
	)

	for i := 0; i < 4; i++ {
		if corners[i].X < xMin {
			xMin = corners[i].X
		} else if corners[i].X > xMax {
			xMax = corners[i].X
		}
		if corners[i].Y < yMin {
			yMin = corners[i].Y
		}
		if corners[i].Y > yMax {
			yMax = corners[i].Y
		}
	}

	return engo.AABB{engo.Point{xMin, yMin}, engo.Point{xMax, yMax}}
}`

(and apologies if this is a bit old, but I think I have the latest)

I just wanted to flag - the logic for updating xMin, yMin, xMax, and yMax is such that they'll never update. corners[i].X will never be less than negative max float. corners[i].Y will neer be greater than max float. etc. The fix, I believe, is that we need to flip the negative signs on the initial assignments to xMin, xMax, yMin, and yMax.

Thoughts?

jamesevans:
flipped the signs on collision.go initial bounds for AABB(), so that AABB() returns the actual max bounds of the Space Component, rather than the max bounds of +/- MaxFloat32
fixes #441

### DIFF
--- a/common/collision.go
+++ b/common/collision.go
@@ -52,10 +52,10 @@ func (sc SpaceComponent) AABB() engo.AABB {
 	corners := sc.Corners()
 
 	var (
-		xMin float32 = -math.MaxFloat32
-		xMax float32 = math.MaxFloat32
-		yMin float32 = -math.MaxFloat32
-		yMax float32 = math.MaxFloat32
+		xMin float32 = math.MaxFloat32
+		xMax float32 = -math.MaxFloat32
+		yMin float32 = math.MaxFloat32
+		yMax float32 = -math.MaxFloat32
 	)
 
 	for i := 0; i < 4; i++ {


paked:
Thanks!

jamesevans:
No problem, happy to help!

Issue #447, opened 2017-07-22
App closing in Android


Noofbiz:
Hello,

In android if you build a game, then close it using the home button, if you try to go back on the game using either the icon or the app switcher without force closing it first, the game crashes and won't reload until you force close it. It's due to the GL context not being destroyed on the app crossofff. I've got it worked out on my bransh so I'm submitting a PR for it.

Thanks!

otraore:
Closed by https://github.com/EngoEngine/engo/pull/448

PR #448, opened 2017-07-22
close game for gomobile

Noofbiz:
Needed to clean up Gl context when the game closes. Otherwise the game just freezes and crashes on next restart unless it's force closed.
We also want to use the device's lifecycle and cross-off when closing rather than use a break to get out of the main loop.

#447


otraore:
Thanks for the contribution!

Issue #428, opened 2017-03-07
Mouse coordinates are off when using gopherjs


countcb:
Can be reproduced with the mouse-follow demo or the mouse demo

My screenshots are not showing the cursor... so no screenshots. :(

PR #450, opened 2017-07-27
Android and gopherjs resize drawing fix

Noofbiz:
Issues #449 and #428 

The problem was actually due to window resizing. You could make it show up on glfw by changing the window size manually while running. When a window resizes, the space components position is still in the original coordinates, so the entities are being drawn to the screen in the old space instead of new. To fix this, I added resize offset variables that keep track of how much the window size changes and it adjusts the position in the shader matrix when drawing to the screen. The issue was compounded in gopherjs due to the canvas scale not being one, which created even more of an offset.

### DIFF
--- a/common/render_shaders.go
+++ b/common/render_shaders.go
@@ -228,8 +228,8 @@ func (s *basicShader) Draw(ren *RenderComponent, space *SpaceComponent) {
 		s.modelMatrix[4] = ren.Scale.Y
 	}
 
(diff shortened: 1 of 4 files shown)


Noofbiz:
I'm going to close this and try fixing it via just the engo_mobile and_js files. Less of a chance for me to conflict with something else somehow. I don't really want to mess with the shader if I don't have to.

PR #451, opened 2017-07-29
Mouse cursor offsets

Noofbiz:
Fixed the mouse cursor offsets in Android (#449) and JS (#428) without having to edit the shader stuff and worry about camera / window / canvas interactions being all messed up by it.

The code used for testing can be found here
https://github.com/Noofbiz/MouseTests

but is just a mashup of several demos I thought would be the most affected. Ran the demos on their own as well, looks good from my end.

### DIFF
--- a/axis.go
+++ b/axis.go
@@ -80,11 +80,11 @@ func (am *AxisMouse) Value() float32 {
 	var diff float32
 
 	if am.direction == AxisMouseHori {
-		diff = Input.Mouse.X - am.old
-		am.old = Input.Mouse.X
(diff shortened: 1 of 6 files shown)


paked:
I don't currently have time to go through and test this. So I'm going to commit the cardinal sin of OSS and just merge it.

@Noofbiz: If you plan on working more on/with Engo in the future, would you like to get maintainer access so you can push/pull without relying on me?

Noofbiz:
That sounds great, I would love to help out as much as I can.

Issue #453, opened 2017-08-05
Font cutoffs


Noofbiz:
For some .ttf files, the font is cutting off before the whole font is rendered in the y-direction.

I have a gist of the issue here:
https://gist.github.com/Noofbiz/aea29d23b8bc5b1f1375a6e7c960f22c

The Roboto font was taken from the font demo, and the Oswald font can be found at google fonts:
https://fonts.google.com/specimen/Oswald

![fontissue](https://user-images.githubusercontent.com/5117030/28993309-1e0b372a-7968-11e7-91be-f96ee66256dd.PNG)

PR #454, opened 2017-08-05
fix font issue

Noofbiz:
The NRGBA image was only as high as the pt size, not the actual height of the font rendered, which is the yBearing size.

Fix for #453


Noofbiz:
Closing to use newer changes that also account for font width as well

PR #455, opened 2017-08-05
common: add whitespace to package comments for better godoc

jrmiller82:
I added whitespace to the package comment in common/interfaces.go which fixed ugly godoc html rendering. It now makes Getters, Faces, etc. appear as nice headings in the godoc Overview section. 

The line 1 `interfaces.go is intended...` looks a bit out of place now in the documentation and should likely be moved around. But, I'm still learning engo so I probably shouldn't be the one to decide that.

### DIFF
--- a/common/interfaces.go
+++ b/common/interfaces.go
@@ -1,12 +1,20 @@
 // interfaces.go is intended to provide a simple means of adding components to each system
+//
 // Getters
+//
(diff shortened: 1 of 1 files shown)


jrmiller82:
Actually going to close this PR and add terminal punctuation to sentences. I'll resubmit shortly.

PR #457, opened 2017-08-05
add documentation to MinZoom and MaxZoom in common/camera.go

jrmiller82:
I spent several minutes tweaking MinZoom thinking it would let me zoom out farther, and lo and behold, it's the opposite! Added some documentation so future people won't have the same issue and can realize that you actually increase MaxZoom in order to zoom "out".


Noofbiz:
I'm restarting the travis-ci on this to see if it hits the same problem as my latest PR in order to find out if it's travis-ci or something in my PR causing it. This shouldn't hit any problems because it's all comments :P

PR #459, opened 2017-08-06
Add "NotResizable" RunOpts option

joegilley:
This adds a `NotResizable` option that hints to `GLFW` that the window
should not be resizable if specified true.  `GLFW` defaults resizable
to true, so `RunOpts` does as well.

With `NotResizable: true` specified:
image


Noofbiz:
In your comment for the option you say it defaults to true, but it actually defaults to false the way it is now. Is that a typo or do you want it to default as true?

Noofbiz:
Awesome, thanks for the addition!

PR #475, opened 2017-11-28
common: parse gid from tmx objects

mischief:
ran go fmt as well, which caused the import/comment shuffling.


mischief:
in the future, it might be desirable to create Drawable wrappers around tmx objects like is done for tmx tiles right now, but that is a bigger change so i have left it for later.

Noofbiz:

PR #479, opened 2017-12-03
center changes

Noofbiz:
Now the center gotten and set from center in the space component adjusts for rotation properly

### DIFF
--- a/common/collision.go
+++ b/common/collision.go
@@ -21,6 +21,14 @@ func (sc *SpaceComponent) SetCenter(p engo.Point) {
 	xDelta := sc.Width / 2
 	yDelta := sc.Height / 2
 	// update position according to point being used as our center
+	if sc.Rotation == 0 {
+		sc.Position.X = p.X - xDelta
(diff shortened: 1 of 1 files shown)


harare:
Hello there. I was happy to hear about the center rotation. But for me it is not working. I tried it with my own code and also with your demo implementation of "rotation". Even if is set the center of a spaceComponent, the rotation point is still the top left corner.

Noofbiz:
Have you tried `go get -u engo.io/engo`?

It isn't working if you get the center then set it? For example, if e is the entity you're trying to rotate and angle is the rotation angle:
```go
center := e.Center()
e.Rotation = angle
e.SetCenter(center)
```
you have to do all that because it will always rotate about the top left corner, but by getting where the center was before the rotation and setting the center back after the rotation, it will appear as though it is rotating about the center. Let me know if that doesn't work for you. Thanks!

harare:
Oh nice. Thanks for your fast support. I already had the newest sources when I wrote my last message. But your second hint was the right solution. Many thanks. Nice work btw!

PR #481, opened 2017-12-13
added support for asymmetric sprite sheets as well as for borders bet…

### LINKED ISSUE #478: Asymmetric sprite shets
According to the site's animation tutorial, Engo has support for asymmetric sprite sheets, but I don't see it anywhere. If there isn't support I would like to implement it

Noofbiz:
…ween sprites on the sprite sheet

Added support for asymmetric sprite sheets. Created a sprite struct that keeps the position and size information about each sprite. Since the symmetric is just a subset of asymmetric, all the symmetric stuff produces sprites on it's own then feeds that to the asymmetric.
Also implemented symmetric sprite sheets with borders between the sprites, which helps with issues that sprites bleed into eachother.
Closes #478

### DIFF
--- a/common/spritesheet.go
+++ b/common/spritesheet.go
@@ -4,23 +4,58 @@ import (
 	"log"
 
 	"engo.io/engo"
+	"engo.io/engo/math"
 	"engo.io/gl"
 )
 
 // Spritesheet is a class that stores a set of tiles from a file, used by tilemaps and animations
 type Spritesheet struct {
-	texture               *gl.Texture     // The original texture
-	width, height         float32         // The dimensions of the total texture
-	cellWidth, cellHeight int             // The dimensions of the cells
-	cache                 map[int]Texture // The cell cache cells
+	texture       *gl.Texture     // The original texture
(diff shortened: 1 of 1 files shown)


common/spritesheet.go
```diff
@@ -4,23 +4,58 @@ import (
...
+	cache         map[int]Texture // The cell cache cells
 }
 
-func NewSpritesheetFromTexture(tr *TextureResource, cellWidth, cellHeight int) *Spritesheet {
-	return &Spritesheet{texture: tr.Texture,
-		width: tr.Width, height: tr.Height,
-		cellWidth: cellWidth, cellHeight: cellHeight,
-		cache: make(map[int]Texture),
+// Sprite holds the position data for each sprite on the sheet
+type Sprite struct {
```
paked: Maybe it'd be better to call this a 'TextureLocation' or 'SpriteLocation' (or maybe 'XRegion')

Noofbiz: Updated to SpriteRegion. Thanks, that is much more clear ^_^

Issue #484, opened 2018-01-07
Type "tile" should be exported


gabre:
I wanted to procedurally create TileLayers. there is a field there called Tiles (of type *common.tile)... the type "tile" is unexported so I am unable to create it. Is there no way to export it?

Noofbiz:
@paked @EtienneBruines @otraore  does anyone see a reason to keep this unexported? It can be handy for procedurally generated games to use TMX instead of just tiles, since TMX does keep track of collisions and generally makes tiling and indexing easier than just using a spritesheet and randomly picking indexes off of that.

gabre:
Exporting function `createTileset` would also be helpful if one wants to create those tiles on his own. (see usage at tmx_level.go:344, definition: level.go:169). Not "a must" but would not be bad to have some function for index conversions at hand (level.go:218, from X,Y to GID)

paked:
Sounds fine to me.

gabre:
@paked one thing: as I see, to use `createTileset` (after it is exported) one needs the tile sheets. This is an info only available when someone has the XML representation of the map unmarshalled. (see tmx_level.go:344) Is it possible to add this to Level (or somehow make it available)?

My use case would be the following: I use an empty TMX file where all the needed tilesets are added to initialize a struct `Level` then procedurally generate the specific tiles for each (X, Y) coordinate. This means that with the current layout of `Level` I don't have the tile sheets after loading the initial, empty map. Loading once again as XML (unmarshalling) just to get the tile sheets (to be able to use them then for tile generation) would mean a little bit of overhead.

EtienneBruines:
@gabre This sounds like a valid use-case. I don't know about @paked or @otraore , but I'm currently probably too busy to 'fully' implement this. 

Are you able to make the changes yourself, exporting what you think you might need / might be useful, and creating a PR for that? We usually merge nice PRs within a day or two. 

Thanks,
Etienne

gabre:
@EtienneBruines I think I can push a PR if you can review it. The only thing that is not clear whether I should put `tile sheets` into the `Level` struct or not. Currently it has no tile sheets so this could have some performance/memory implications (but not big).

PR #485, opened 2018-01-08
Export tiles and tilesets

gabre:


common/tmx_level.go
```diff
@@ -72,24 +72,24 @@ type TMXData struct {
 	// valid values are one of "", "zlib", "gzip"
 	Compression string `xml:"compression,attr"`
 	// Data contains an encoded list of uint32 guids defining
-	// the tile layout of a layer
+	// the Tile layout of a layer
 	Data string `xml:",innerxml,"`
 	// Tiles is an array of tiles containing guids. Not set if
 	// other encodings are used
-	Tiles []TMXTile `xml:tile`
```
Noofbiz: Don't change the xml parsing comment to be capitalized, that may mess up unmarshalling

gabre: Ah that was Goland's super renaming refactoring functionality. Sorry, one has to do everything manually.

Noofbiz: No worries, I only knew to be wary of that because I've done exactly that myself before lol

EtienneBruines:
We may want to reference that this fixes #484

gabre:
I ran a manual test:
- first, I ran TMX demo from this PR, it worked fine (note that I changed TMX demo's map file)
- then, I cherry-picked the modified map file onto master where it crashed (panic: runtime error: index out of range)

Noofbiz:
Looks good. Just ran all the demos that use tile maps in some capacity, and all of them are the same, as expected. Thank you for your contribution!

PR #487, opened 2018-01-13
Allow TMX parser to use tile size defined in tilesheet

gabre:
Signed-off-by: gabre <gabre@caesar.elte.hu>

### DIFF
--- a/common/level.go
+++ b/common/level.go
@@ -159,6 +159,8 @@ type Tile struct {
 type tilesheet struct {
 	Image    *TextureResource
 	Firstgid int
+	Width    int
+	Height   int
 }
 
 type layer struct {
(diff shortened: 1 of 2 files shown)


common/level.go
```diff
@@ -170,10 +172,14 @@ type layer struct {
 
 func createTileset(lvl *Level, sheets []*tilesheet) map[int]*Tile {
```
gabre: Should we somehow expose the tilesheet's tile width/height or are we OK with only having them in the tiles (in the map[int]*Tile) as now?

Noofbiz: Do you need the tiles exposed or can you get by with the map?

gabre: With this modification (as is) I seem to get by (at least now :) ).

PR #495, opened 2018-02-07
safety on nil added to dispatch

coderconvoy:
The purpose of this pull request is to make it easier to write tests on systems that dispatch messages.

This way the system can be invoked and still call Engo.Mailbox.Dispatch, even if it hasn't been created by the runtime.

This can be worked around, but the reason I thought it would be worth making as a pull request is that we want to encourage testing, and taking away a hurdle to that could be helpful

### DIFF
--- a/message.go
+++ b/message.go
@@ -15,6 +15,10 @@ type MessageManager struct {
 
 // Dispatch sends a message to all subscribed handlers of the message's type
 func (mm *MessageManager) Dispatch(message Message) {
+	if mm == nil {
+		//systems that call dispatch are easier to test without invoking the whole system
+		return
+	}
 	handlers := mm.listeners[message.Type()]
 
 	for _, handler := range handlers {


Noofbiz:
Wouldn't this throw a nil pointer reference when you do mm.Dispatch() and never make it to that point? To test systems that dispatch messages wouldn't it be best to create an actual message manager and then a test system that listens for the messages, so you know it is being sent and able to be received by other systems?

coderconvoy:
Oddly this works here. Unlike most languages I've seen, the function can call on a type even if it's nil, but obviously you can't reference any internal data.

It's possibly not the best way to handle it. I pondered getting it to log or something since it was for test purposes. But could'nt see an easy way round. 

I also considered having the system recieve a Dispatcher as a parameter, (dependency injection) but I thought that was too much work. 

In some recent work I actually did

    if engo.Mailbox != nil {
       engo.Mailbox.Dispatch(mess)
    }

which worked for my case, and I could test it safely.

It was just a thought that the mailbox could do this and save a test case from outside.

Noofbiz:
I just mean wouldn't it be okay if you just made your own message manager instead of engo.Mailbox for testing. That way you get (and can test) the sending and receiving of messages without the overhead of the entire engo engine

coderconvoy:
Agred,

PR #496, opened 2018-02-07
HitSystem - An interface powered CollisionSystem Demo

coderconvoy:
The purpose of this pull request is to intorduce the idea of using interfaces to the engo.io system in the way that golang does best.

The Hit System introduced is a demo of what the Collision System could be if it used interfaces to the full.

Engo.io implements the ECS architecture where each component does not act upon itself, and so in order to use the components effectively, we have to split them up and create new objects based on fragments of them in order to keep the type system together.

My suggestion is a slightly more relaxed model, wherin the systems work on interfaces not sets of components. We allow Components a few small methods that work upon themselves and can be combined by anonymous structs on order to achieve larger interfaces.

The complex and interactive work remains the systems job, but minor adjustments, can be done as calls to an object itslelf.

Benefits of this system
--------------------------

Components can be tailored more directly to meet a systems requirements without using any of a set of different components that fulfil the required interfces instead of being forced into systems they dont fit.

It is possible for the Render System, and Collision System, to use different area boxes, (My primary motivation for making this change)

Systems only need to ask for the few methods they need, rather than having to require a large set of un-needed things.

### DIFF
--- a/common/hitsystem.go
+++ b/common/hitsystem.go
@@ -7,10 +7,13 @@ import (
 
 type HitGroup byte
 
+//HitBox is currently a simple bounds rect.
+//I hope to add a possibility of internal rects, for more fine grain hit tests at some point
 type HitBox struct {
 	x, y, w, h float32
 }
 
+//Hit tests whether this Hitbox, has an area collision with another HitBox
 func (a HitBox) Hit(b HitBox) bool {
 	if a.x > b.x+b.w {
 		return false
@@ -27,9 +30,14 @@ func (a HitBox) Hit(b HitBox) bool {
 	return true
 }
 
-//Minimum step a needs to take to get off of B
+//MinimumStepOffD calculates the mininum step this HitBox needs to take to get off of 'b'
+//Return dx, dy
 func (a HitBox) MinimumStepOffD(b HitBox) (float32, float32) {
-	angle := 0 // top
+	//angle and dist are the current winners
+	//0 = up, 1 = right ...
+
+	// top
+	angle := 0
 	dist := a.y + a.h - b.y
 
 	// right
@@ -57,6 +65,8 @@ func (a HitBox) MinimumStepOffD(b HitBox) (float32, float32) {
 
 }
 
+//Hitable is the main interface for the HitSystem
+//Entities can implement the methods by containing Components that implement them
 type Hitable interface {
 	ID() uint64
 	GetHitBox() HitBox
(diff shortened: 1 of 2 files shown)


Noofbiz:
I don't see why not. Add is not defined in the System interface, so you can use whatever method you like there. Are you suggesting adding Add(interface{}) to the System interface as well?

coderconvoy:
No need to change the System interface. It's fine as is. 

In fact putting the Add(interface{}) method on System would be harmful to type safety.


Many "common" systems can actually be changed to fit this way of working without any breaking changes as long as the required components are given appropriate methods.

The main reason I want this is to be able to define a different hitbox than render area. But I think the benefits will be felt everywhere if the whole engine would work to interfaces.

I'm happy to go through a few other systems and tweak them to see what you think. But I didn't want to do too much without checking first.

Noofbiz:
All the systems in common have something like this with the AddByInterface functions, I don't think they'd need to be updated for you to implement it yourself. We could even do a whole series about doing it this way in the tutorials, but I don't think we need to go through every system in common and change them to this style.

Noofbiz:
One thing you could do to get the functionality you want is by creating a second space component on the entity, pass one space component to the render system and the second space component to the collision system. Then you can create a new system between the render and collision ones that takes in both the space components and makes sure they're equal however you want (centered around the same point for example)

common/hitsystem.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"engo.io/ecs"
+	"engo.io/engo"
+)
+
+type HitGroup byte
+
+//HitBox is currently a simple bounds rect.
+//I hope to add a possibility of internal rects, for more fine grain hit tests at some point
+type HitBox struct {
```
Noofbiz: Couldn't we use an engo.AABB? There's also an engo.Container interface that I think would be a good fit for this.

common/hitsystem.go
```diff
@@ -0,0 +1,155 @@
...
+//Type fulfils the engo Messagebox interface
+func (HitMessage) Type() string {
+	return "HitMessage"
+}
+
+//HitSystem is the replacement system for the CollisionSystem
+//It runs on interfaces instead of directly working on Components
+type HitSystem struct {
+	Solid    HitGroup
+	Entities []Hitable
```
Noofbiz: I think it'll have to be []*Hitable if you want to send the actual entity via message and not just a copy.

common/hitsystem.go
```diff
@@ -0,0 +1,155 @@
...
+			e2main, e2grp := e2.HitGroups()
+			if vmain&e2grp == 0 {
+				continue
+			}
+			ghb := e2.GetHitBox()
+			if !vhb.Hit(ghb) {
+				continue
+			}
+			//is a Hit
+			if engo.Mailbox != nil { // Workaround for safe testing
```
Noofbiz: This could be done by creating a mailbox in the tests, so this is not really necessary here

coderconvoy:
Overall I think your wisdom is right on this. 

I think I'll probably do most of my own systems oninterfaces, but there's no need to change this.

One of the reasons I put in a hitbox type was because I was hoping to somehow extend it to act a bit like a tree. so an object could be hit with multiple bounds.
 
I sstill want to play with that. If I come up with a good plan for that I'll let you know

common/hitsystem.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+}
+
+//Hitable is the main interface for the HitSystem
+//Entities can implement the methods by containing Components that implement them
+type Hitable interface {
+	ID() uint64
+	GetHitBox() HitBox
+	HitGroups() (HitGroup, HitGroup)
```
LasTshaMAN: I think, adding comments for methods like these won't hurt (as it is not clear why there are two hit groups returned).

coderconvoy:
I don't think this is gonna be pulled anyway. It doesn't fit the ecs model at all. And there are other problems.

That said the hitbox groups are for main and group.

Noofbiz:
I actually liked this idea, but we don't really need another collision system. I've actually submitted a PR to ecs that'll enable you to use AddByInterface in all your systems to automatically add entities to `System`s as long as the implement the interface for that system. Hope that'll help you out!

PR #498, opened 2018-02-10
Equal collision

coderconvoy:
The purpose of this pull request is to make collisions between equal objects act equally.

Currently, the object who comes earlier in the list is checked for collision first (for loop, hard to avoid)
This first object is then moved off of the second, and so the second doesn't even register the collision when his turn comes around.

An unfortunate consequence of this is that if both elements are free to move around, the second in the list will always push the first around, but the first cannot push the second.

This recently cuased a lot of confusion for a user on gitter.

To fix this, we ask if the second element also has a main collision with the first object, (only if the collision is solid). If so we move them both back half the escape distance, and dispatch a message for both of them.


Noofbiz:
Thank you for the contribution!

PR #506, opened 2018-02-27
clock unit tests added

Noofbiz:
Added tests to test the Clock. It adds 3 seconds to the total test time, but because the FPS ticker is setup to go once every second, it seems like it's going to have to be this way. Anyone have any suggestions that might speed this up?

### DIFF
--- a/clock.go
+++ b/clock.go
@@ -4,6 +4,22 @@ import (
 	"time"
 )
 
+// timer is a wrapper for time.Now().UnixNano() so we can test
+// this wihout relying on time
+type timer interface {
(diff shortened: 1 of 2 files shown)


hajimehoshi:
In general, a mock (monotonic) timer is preferred for timer testing to an actual timer since an actual timer is unreliable in precision, and as you mentioned, it takes long. Sorry if I miss some context.

Noofbiz:
Thanks! I managed to build a test timer that is controlled by the tests rather than being normal clock. It works nicely.

PR #512, opened 2018-03-06
add separate jobs to travis-ci builds

Noofbiz:
Adding different env variables to the test matrix to split the build into multiple jobs. This should speed things up and actually test if it builds on all systems


Noofbiz:
From 20 minutes to 5, and it tests more than it did before! ^_^

Issue #318, opened 2016-05-19
CameraSystem cleanup


EtienneBruines:
As @TheBerryBeast noted, the `RenderSystem` adds a `CameraSystem` if one isn't present yet. 

However, users could add one _right after_ adding the `RenderSystem` - this will not work (nor will it ever). We should find a way to handle this. Perhaps letting the `RenderSystem` crash if no `CameraSystem` was added before adding the `RenderSystem`? That would require everyone to explicitely add the `CameraSystem`, but will also prevent people from accidentally adding it afterwards, thinking it'll work. 

Any thoughts, workarounds or suggestions are welcome - from anyone.

paked:
I think we could take some inspiration from [negroni](https://github.com/urfave/negroni) and have an "add default systems" sort of function. `common.DefaultSystems(w *World)` or something?

The function would add all of the "recommended systems" in the correct order. This might be worth moving to another issue as it doesn't completely fix this one.

Noofbiz:
I don't see any problem with the `RenderSystem`, or any `System`, automatically adding other systems they require so long as those systems have no entities. The real issue here is attempting to add multple `CameraSystem`s, since a game can only have one camera. I think we should check at `CameraSystem.New()` if there's more than one, and then panic if there is. We should also update the documentation for the `RenderSystem` to say that it adds a `CameraSystem` if one is not present as well as documentation for the `CameraSystem` to say that it panics if more than one `CameraSystem` is added to the `World`. Maybe have the panic mention that `RenderSystem` automatically adds a `CameraSystem` so people will see that and go "oh."

PR #514, opened 2018-03-13
Warning if multiple cameras are added

Noofbiz:
Now panics when more than one `CameraSystem` is added to a `World`. Added documentation to `CameraSystem` to reflect this, and to `RenderSystem` to warn that it adds a `CameraSystem` if one isn't present. The panic also says this to provide a hint for anyone with issues. Also added tests for the panic.

### DIFF
--- a/common/camera.go
+++ b/common/camera.go
@@ -48,7 +48,9 @@ type cameraEntity struct {
 	*SpaceComponent
 }
 
-// CameraSystem is a System that manages the state of the virtual camera.
+// CameraSystem is a System that manages the state of the virtual camera. Only
+// one CameraSystem can be in a World at a time. If more than one CameraSystem
+// is added to the World, it will panic.
 type CameraSystem struct {
 	x, y, z  float32
 	tracking cameraEntity // The entity that is currently being followed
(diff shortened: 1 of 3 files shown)


Noofbiz:
This closes #318

common/camera.go
```diff
@@ -60,7 +62,18 @@ type CameraSystem struct {
...
+func (cam *CameraSystem) New(w *ecs.World) {
+	num := 0
+	for _, sys := range w.Systems() {
+		switch sys.(type) {
+		case *CameraSystem:
+			num++
+		}
+	}
+	if num > 0 { //initalizer is called before added to w.systems
+		panic("More than one CameraSystem was added to the World. The RenderSystem adds a CameraSystem if none exist when it's added.")
```
arxdsilva: Hi @Noofbiz, I dont think that panicking would be nice here (saw many panics in code) and I believe that It'd be better to just return an error so we can handle it later, or in the last case if you want to keep this behavior of `New`, a `log.Fatal(err)` would get rid of the stack trace and return only the error.

Noofbiz: @arxdsilva Upon reflection, I agree. An additional `CameraSystem` isn't going to wreck anything, just take up memory unnecessiarly. I think a `panic` or `log.Fatal(err)` would be going too far. A warning printed to the log without crashing the game should suffice.

PR #515, opened 2018-03-14
Renders the texture based on the space component's size and render component's repeat

### LINKED ISSUE #237: Repeating textures
Let's say we have a texture that's 200x200 pixels. But we want to draw something that's 400x400 pixels. We could create four entities, put them next to each other, and be done with it. But it's better to tell the GPU to "repeat" the texture. 

It has something to do with changing the settings for these lines of code:

``` go
Gl.TexParameteri(Gl.TEXTURE_2D, Gl.TEXTURE_WRAP_S, Gl.CLAMP_TO_EDGE)
Gl.TexParameteri(Gl.TEXTURE_2D, Gl.TEXTURE_WRAP_T, Gl.CLAMP_TO_EDGE)
Gl.TexParameteri(Gl.TEXTURE_2D, Gl.TEXTURE_MIN_FILTER, Gl.LINEAR)
Gl.TexParameteri(Gl.TEXTURE_2D, Gl.TEXTURE_MAG_FILTER, Gl.NEAREST)
```

It could be they're already set correctly, but then we should allow people to actually use it. 

---
## How

We can look at the width/height as defined in the `SpaceComponent` to get an idea of how large the texture should be. The tricky part, is telling the shaders that. Currently [these lines](https://github.com/EngoEngine/engo/blob/master/render.go#L111-L112) tell the GPU how large the texture should be. If we were to update those, we'd be done. 

The tricky part is, that these lines of code are within the `RenderComponent` itself, without having any reference to a `SpaceComponent`. We could (and should!) move that function to be shader-specific anyway, which would allow us to have some kind of reference to some kind of `SpaceComponent`. 

Another reason for moving it to the shaders, is because the buffer-contents (static stuff about the texture, as stored in the GPU RAM) may change from shader to shader (one shader might require other values to be present). 

Another tricky part, is that this method is currently being cached, heavily. Because it updates the GPU RAM, changing it is very resource-expensive. So far, it's only being updated whenever values change (that's why we have a `SetScale` function, and a `SetDrawable` function, to name a few). However, if this were moved to the shaders, then we'd have to figure out when to update those buffers. 

Another thing to note, is that not everyone wants to repeat his texture. Some only want to repeat the outer pixels (I believe that's what `CLAMP_TO_EDGE` does). We should allow people to set this **per-texture**. 
## Proposal
- Move [this method](https://github.com/EngoEngine/engo/blob/master/render.go#L88) to a per-shader method. 
- Change [calls like these](https://github.com/EngoEngine/engo/blob/master/render.go#L70) to some kind of "hasChanged" setting, to indicate the per-shader method has to be called upon rendering. (This is also faster! :smile: )
- Allow people to set the `TEXTURE_WRAP` (repeating/not-repeating) **per texture**. I believe we could choose some kind of default (don't really care which, by default people won't care) - and have some kind of `TextureWrap` variable on the `RenderComponent`. 
- As an additional note, if we have a `TextureWrap` variable, we might as well include it in [the sorting definition](https://github.com/EngoEngine/engo/blob/master/render.go#L196-L204). This way we don't have to set different values all the time. (meaning less GPU calls)

Noofbiz:
This allows the RenderComponent's Repeat to work properly. It's based on the size of the SpaceComponent relative to the size of the RenderComponent's Drawable * Scale. This closes #237 

It adds a new TextureRepeat as the default, NoRepeat. If this is selected (or none are), it will draw the texture as if the View() were 0,0,1,1; just like it did before. 

However, if another one is chosen it will now look at the SpaceComponent's size to get how big it should draw the the screen, as well as the results for the View. If the SpaceComponent is smaller, it'll only draw a portion of the texture; if it's larger it'll repeat the texture based on the selected Repeat.

### DIFF
--- a/common/render_shaders.go
+++ b/common/render_shaders.go
@@ -209,6 +209,8 @@ func (s *basicShader) Draw(ren *RenderComponent, space *SpaceComponent) {
 	if s.lastRepeating != ren.Repeat {
 		var val int
 		switch ren.Repeat {
+		case NoRepeat:
+			val = engo.Gl.CLAMP_TO_EDGE
 		case ClampToEdge:
 			val = engo.Gl.CLAMP_TO_EDGE
 		case ClampToBorder:
@@ -285,7 +287,16 @@ func (s *basicShader) generateBufferContent(ren *RenderComponent, space *SpaceCo
 
 	tint := colorToFloat32(ren.Color)
 
-	u, v, u2, v2 := ren.Drawable.View()
+	var u, v float32
+	u2 := float32(1)
+	v2 := float32(1)
+
+	if ren.Repeat != NoRepeat {
+		u2 = space.Width / (ren.Drawable.Width() * ren.Scale.X)
+		w *= u2
+		v2 = space.Width / (ren.Drawable.Height() * ren.Scale.Y)
+		h *= v2
+	}
 
 	var changed bool
 
--- a/common/render.go
+++ b/common/render.go
@@ -36,8 +36,10 @@ type Drawable interface {
 type TextureRepeating uint8
 
 const (
+	// NoRepeat does not repeat the texture.
+	NoRepeat TextureRepeating = iota
 	// ClampToEdge stretches the texture to the edge of the viewport.
-	ClampToEdge TextureRepeating = iota
+	ClampToEdge
 	// ClampToBorder stretches the texture to the border of the viewpport.
 	ClampToBorder
 	// Repeat repeats the texture until the border of the viewport.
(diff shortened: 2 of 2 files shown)


common/render_shaders.go
```diff
@@ -285,7 +287,18 @@ func (s *basicShader) generateBufferContent(ren *RenderComponent, space *SpaceCo
 
 	tint := colorToFloat32(ren.Color)
 
-	u, v, u2, v2 := ren.Drawable.View()
+	//u, v, u2, v2 := ren.Drawable.View()
+	u := float32(0)
+	v := float32(0)
```
arxdsilva: you can assign `u` and `v` to something like this: `var u,v float32` (https://play.golang.org/p/LnCUPb50j-4)

Noofbiz:
Fixed ^_^ Thanks

PR #519, opened 2018-03-20
choose between nearest neighbor and linear filtering in RenderComponent

Noofbiz:


arxdsilva:
hi @Noofbiz , could you add some tests to this? Just to make sure that It's 100% as you want :D

Noofbiz:
Yeah, I've been thinking of ways to test shaders and rendering in general without *actually* making ogl calls, since I'd like all the tests to be headless and automated (not require a user to go "yes, this picture is better than that one!"), but also I'll assume external libraries such as glfw are tested in their own right. 

I'm thinking of making engo.Gl an interface instead of a *gl.Context (which *gl.Context implements), and then creating a TestGl struct that also implements that interface.That way we can test shaders and rendering in general without having to actually call ogl unless necessiary, such as testing that the shaders compile in glsl

arxdsilva:
I'll see what I can do about testing, mainly in common so we can get assertive

PR #562, opened 2018-06-08
fixed duplicate movement in CollisionSystem

qutrace:
Bug:
When a CollisionEntity with the Main field set is intersecting two or more other entities, (for example, when it's pushed into the ground by gravity where two ground tiles meet) it is affected by both Minimal Translations, when one would be sufficient.
Fix:
After applying the first MinimalTranslation the new position of the CollisionEntity is used as further reference to all other isIntersecting checks


Noofbiz:
Thank you for your contribution! Looks great to me!

Issue #559, opened 2018-05-31
engo.Point methods return the point


Noofbiz:
engo.Point methods could return an engo.Point, so you can chain together multiple calls:

```go
p1 := engo.Point{X: 5, Y:5}
p2 := engo.Point{X: 2, Y:3}
p1.Add(p2).MultiplyScalar(2).Normalize()
```

jeet-parekh:
Seems like a good first issue. Is it still just a proposal? I'd like to give it a shot, if you're not working on it @Noofbiz.

PR #563, opened 2018-07-23
engo.Point methods now return the point

jeet-parekh:
Implements #559.


Noofbiz:
Don't worry about the travis CI build,it looks like it's a new dependency from oto. I'll get that fixed soon and then get this in! Thanks for the contribution!

Noofbiz:
Thanks for your contribution!

jeet-parekh:
Glad to be a part of it!

Issue #575, opened 2018-09-25
Race conditions on Headless when closing via Terminal


edoardo849:
I'm detecting race conditions when a game is closed via the Terminal and not via the window. For example:

```bash
git clone git@github.com:EngoEngine/engo.git
cd engo/demos/pong
go build -race pong.go
./pong

2018/09/25 15:06:51 The score is now 1 vs 0
2018/09/25 15:06:52 The score is now 2 vs 0
2018/09/25 15:06:52 The score is now 3 vs 0

# I press CTRL+C
$ ^C
==================
WARNING: DATA RACE
Read at 0x00c000098210 by goroutine 8:
  runtime.mapiterinit()
      /usr/local/go/src/runtime/map.go:734 +0x0
  engo.io/engo.closeEvent()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo.go:270 +0x94
  engo.io/engo.runLoop.func1()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo_glfw.go:259 +0x46

Previous write at 0x00c000098210 by main goroutine:
  [failed to restore the stack]

Goroutine 8 (running) created at:
  engo.io/engo.runLoop()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo_glfw.go:257 +0x128
  engo.io/engo.Run()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo.go:217 +0x8c2
  main.main()
      /Users/escalafiotti/code/go/src/github.com/edoardo849/engo/demos/pong/pong.go:455 +0x137
==================
==================
WARNING: DATA RACE
Read at 0x00c00009aa48 by goroutine 8:
  engo.io/engo.closeEvent()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo.go:270 +0xc5
  engo.io/engo.runLoop.func1()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo_glfw.go:259 +0x46

Previous write at 0x00c00009aa48 by main goroutine:
  [failed to restore the stack]

Goroutine 8 (running) created at:
  engo.io/engo.runLoop()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo_glfw.go:257 +0x128
  engo.io/engo.Run()
      /Users/escalafiotti/code/go/src/engo.io/engo/engo.go:217 +0x8c2
  main.main()
[...]

PR #576, opened 2018-09-25
Fixes race conditions 

edoardo849:
Fixes issue #575 : when running a game and closing it via Terminal via `os.Interrupt` or `syscall.SIGTERM` (aka CTRL+C) 8 race conditions are detected if the game is built or run with the `-race` flag


Noofbiz:

Issue #557, opened 2018-05-31
Splitting Common into multiple packages


Noofbiz:
Or maybe even it's own separate repo, like "engo.io/systems"

This would be useful for anyone trying to write their own systems, they can pick and choose which ones are added and which aren't, rather than importing common and getting all of them. Especially helpful because the init for filetypes adds things like how engo handles .png, for example. If you want to use your own render system and open .pngs differently, you'd have to work around the fact that common already registers that filetype. It would also be nice pretty cool to call, for example, the RenderSystem via `render.System` and RenderComponent via `render.Component`.

Noofbiz:
So the issue I'm having is we can either do stuttering, i.e. `render.RenderSystem`, or we can do `render.System` but the downside to that is namespace collision inside the entities, like if you do 

```
type guy struct {
    ecs.BasicEntity
    space.Component
    render.Component
}
```

It won't compile because it has two `Component`s in one struct. However if you do

```
type guy struct{
    ecs.BasicEntity
    SpaceComponent space.Component
    RenderComponent render.Component
}
```

then guy no longer satisfies the RenderAble interface as guy has no method `GetSpaceComponent() *SpaceComponenet` or `GetRenderComponent() *RenderComponent` functions, guy.SpaceComponent/guy.RenderComponent does!

I'm leaning toward just ignoring the stuttering, as that makes it easiest for end users. Anyone have any ideas/comments on this?

paked:
Take my opinion with a bit of a grain of salt as I'm not really involved here much anymore, but:

I think splitting Engo into common/not-common was a mistake in the first place. We should have left them in one package. I don't think "including all the systems" is too much of an issue for people to deal with, as long as they don't tax the application at runtime it's probably OK. Hell, the compiler may even straight up cut them out of the build.

I'd vote to move everything back into the same package, but work on really decoupling them from the inner codebase (ie. find ways to allow custom loaders, if that's a feature you want to pursue).

Again, grain of salt :)

For reference reading: https://blog.golang.org/package-names

Noofbiz:
After trying this out in hypnic, it turns out to just be a big headache with import cycles, go vet issues, and namespace collisions. I don't think it's worth it. Closing this for now. If there's a reason to do it, we may address it in the future and reopen this.

PR #582, opened 2018-10-06
 #557 Splitting Common into multiple packages

juztme87:
Proof of Concept
Common package was split into packages audio, common2 and render
Additionally math.go was moved from engo base into sub package math2d.

Purpose: Enable servers without GL dependency

### DIFF
--- a/audio/audio.go
+++ b/audio/audio.go
@@ -1,0 +1,150 @@
+package audio
+
+import (
+	"log"
+	"runtime"
+
+	"engo.io/ecs"
+	"engo.io/engo"
+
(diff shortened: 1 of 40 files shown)


juztme87:
Note: I didn't update all demos before we discussed all aspects of this change.

Noofbiz:
Actually, use of the collision system and positioning and everything was not a use case I had thought of when I closed the original issue. That said, if we're going to do it we should do it correctly. I think we should split common into it's own repo entirely like `engo.io/systems`  and then within systems we'd do a separate package for everything like `engo.io/systems/render`

PR #585, opened 2018-10-08
#584 Improved message manager 

### LINKED ISSUE #584: Improved message manager
Add a way to StopListening to methods, maybe even include a wrapper for ListenOnce that'll only listen to the first message received. Also add priorities to messages, so we can control the order the listeners are handled in.

MrTrustworthy:
Added StopListen and ListenOnce to the MessageManager

Taking a first shot to resolve #584 
As this is my first Golang project, please check the code for beginner mistakes :)

What I'm currently not sure about: Locking
Dispatch() only holds a RLock when dispatching the individual handlers. However, StopListen() might (and ListenOnce certainly WILL) modify the handlers list while Dispatch() is iterating over it. Not entirely sure, but this seems very problematic. 

Potential solution: Have Dispatch() use Lock() instead of RLock(). Have StopListen() test the lock. If the lock is open, we can simply remove the handler without special care. If it's already locked (indicating that Dispatch() is currently running), add the MessageHandlerID to some "to be deleted" list. After processing handlers, Dispatch() goes ahead and removes all handlers in the "to be deleted" list. 
Would an approach like that work? Are there cases where StopListen() might run into a lock, even if it's not originating from Dispatch()?

### DIFF
--- a/message.go
+++ b/message.go
@@ -1,7 +1,6 @@
 package engo
 
 import (
-	"errors"
 	"sync"
 )
 
@@ -37,12 +36,14 @@ type MessageManager struct {
 	// conditions on listeners and
 	// sync its state across the game
 	sync.RWMutex
-	listeners map[string][]HandlerIDPair
+	listeners        map[string][]HandlerIDPair
+	handlersToRemove map[string][]MessageHandlerId
 }
 
 // Dispatch sends a message to all subscribed handlers of the message's type
 func (mm *MessageManager) Dispatch(message Message) {
 	mm.Lock()
+	mm.clearRemovedHandlers()
 	handlers := mm.listeners[message.Type()]
 	mm.Unlock()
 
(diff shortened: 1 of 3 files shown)


MrTrustworthy:
Updated PR with a good solution for the locking issue after input from @Noofbiz

message.go
```diff
@@ -67,29 +68,46 @@ func (mm *MessageManager) Listen(messageType string, handler MessageHandler) Mes
...
+	mm.handlersToRemove[messageType] = append(mm.handlersToRemove[messageType], handlerId)
+}
+
+// Will deleted all queued handlers that are scheduled for removal due to StopListen()
+func (mm *MessageManager) clearRemovedHandlers() {
+	for messageType, handlerList := range mm.handlersToRemove {
+		for _, handlerId := range handlerList {
+			mm.removeHandler(messageType, handlerId)
+		}
+	}
```
Noofbiz: At the end here clear out the handlersToRemove so you're not iterating over things that have already been removed.
`handlersToRemove = make(map[string][]MessageHandlerId)` will do the trick ^_^

PR #586, opened 2018-10-09
Fix tmx objects not loaded

hismito:


Noofbiz:
Thank you for the contribution!

PR #598, opened 2018-11-04
explicit audio return

Noofbiz:


Noofbiz:
Not sure why exactly, but all the audio isMobile stuff added can be removed just by adding an explicit return at the end of AudioSystem's update

PR #599, opened 2018-11-06
Export the buffer and buffer content

tenshi1989:
In the case of writing an own shader, there is now the possibility to access the buffer / buffer content from the render component


Noofbiz:
Thanks for the contribution!

PR #601, opened 2018-11-07
fixed AABB max / min to return correctly

Noofbiz:


Noofbiz:
Thanks @boombuler for the catch

PR #609, opened 2018-12-03
add panic if shaders don't compile

Noofbiz:
This will allow custom shaders to know if their shaders have errors and don't work

### DIFF
--- a/common/render.go
+++ b/common/render.go
@@ -164,7 +164,9 @@ func (rs *RenderSystem) New(w *ecs.World) {
 	addCameraSystemOnce(w)
 
 	if !engo.Headless() {
-		initShaders(w)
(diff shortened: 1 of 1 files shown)


common/render.go
```diff
@@ -164,7 +164,10 @@ func (rs *RenderSystem) New(w *ecs.World) {
 	addCameraSystemOnce(w)
 
 	if !engo.Headless() {
-		initShaders(w)
+		err := initShaders(w)
+		if err != nil {
+			panic(err)
+		}
```
hajimehoshi:
```go
if err := initShaders(w); err != nil {
    panic(err)
}
```

maybe?

Noofbiz: Nice catch! Thanks ^_^

PR #612, opened 2018-12-06
Fix race condition in message dispatching.

jonbonazza:
There was a race condition during message dispatching that could
potentially cause a panic (and thus a crash) if the list of message
handlers was modified while a message was actively being dispatched.

The solution here is to make a copy of the the handler list before
iterating over it.


Noofbiz:
Looks good! Thanks for the contribution!

PR #613, opened 2018-12-06
Fix several linting issues that are backwards compatible.

jonbonazza:
Unfortunately, there are still two linting issues that cannot be fixed without breaking backwards compatibility. :(


Noofbiz:
Awesome, thanks!

PR #614, opened 2018-12-06
Use a channel for closeGame.

jonbonazza:
Instead of a RWMutex and a bool, use a channel for signaling when the
game should exit. This is less error prone, more conventional, and
an ultimately clearner implementation.


Noofbiz:
Thank you for your help!

PR #615, opened 2018-12-06
Add bash script for running tests

jonbonazza:


Noofbiz:
I actually like this idea, but I'm not sure it would go in the repo itself. Maybe we add a section to the website that is for engo development. We could add this and other things like an archetecture diagram of engo and the associated systems and ecs libraries.

jonbonazza:
@Noofbiz there is one already for windows (runtests.bat) but i was testing on linux and wrote one for that. Thought maybe others would find it useful as well.

Noofbiz:
Ah, yeah, that's for Appveyor. There's another one for TravisCI `.runtests.sh` I still think having them explicitly somewhere for people to look at would be a great idea

jonbonazza:
@Noofbiz I disagree. Developers need a convenient way to test their changes and without a script like this it's actually quite troublesome to do given the number of demos there are.

Would it be better if we made a makefile instead?

Noofbiz:
I agree with you on that point, I was just suggesting that we make a development section of the website and put it there. I say that because everyone is going to have a different development setup. For example, I have a folder ~/scripts on my $PATH where I keep a similar script but also an executable to run it so all I have to do is run 'engoTestBuild args' from anywhere and it'll test and then make sure the demos build with go, gomobile, and gopherjs depending on the args.

Noofbiz:
I put a copy of the script I use here 

https://gist.github.com/Noofbiz/c369a8093f4c0e90f25803ed92401abf

What I do is compile that to a binary then put it in ~/scripts and I can just call it via `engoTestBuild -gopherjs -gomobile` and it'll try to build native, js, and mobile and panics if there's any failures.
I just never really thought about it as something people would want, honestly. Just as something for the CI builds to pick up.

jonbonazza:
Closing this as no longer needed.

PR #617, opened 2018-12-09
Add sprite batching to basicShader.

jonbonazza:
I tried to do this as backwards compatible as possible. I think the only
place where we potentially lose backwards compatibility is the
basicShader.SetCamera method. I modified it slightly so that a change in
camera would properly recreate the view matrix.


jonbonazza:
I still need to add unit tests for the new Matrix type, but this should let you get a head start on finding anything you don't like.

Also, the batch size is currently exposed as a `basicShader.BatchSize` but since there is no way currently to set this from outside the `common` package, it will always be set to the default `MaxBatchSize`. We should try to come up with a good interface for changing this.

Noofbiz:
Looks good to me. I'm going to try benchmarking it against the old way to see what improvements we get ^_^ I'll let you know how it goes!

Noofbiz:
I'm using this to run benchmarks if you wanna try yourself

https://gist.github.com/Noofbiz/af17111cc0cc197c6bd631134b9d02e0

Noofbiz:
If I add via the sprite sheets, the current way drops below 60fps around 5,000 sprites. If I add only from the sprite sheet on this branch I get above 60fps even at 20,000 sprites, so that's awesome!
However, if I add separate textures, the new way drops below 60fps with less than 500 sprites drawn. With the current way it can draw 3,000 before dropping.

jonbonazza:
Hmm interesting... In the new way, if the texture changes it does flush
before moving on, which is inevitable, but i wouldnt expect it to be any
worse than what we currently have. There must be some inefficiency in
either the flushing or resetting state after flush... Ill need to
investigate. Anything seem out of place there to you?

Oh! I think I know what is wrong. If the texture changes and so do the
texture properties, its going to flush like 5 times (dont have code in
front of me atm). I bet thats it.

Do we need to flush if the texture properties changed? I think we can get
away without.

Noofbiz:
I think if you order the render entities by zindex but then by texture, then it wouldn't change so often. Rather than between each entity only between textures.

jonbonazza:
@Noofbiz So I just tried that out. In your benchmark, every sprite is different, so it doesn't really help. I'm still confused why performance would be any different here vs the original implementation.

Also, FWIW, removing the extra flushes in the basicShader's Draw method helped a little (roughly 2x performance) but it's still not great.

Noofbiz:
Every sprite is different, but there's only 10 images it's drawing from. if you can sort by *gl.Texture then all the ones drawing from one image should have the same underlying *gl.Texture. If the *gl.Texture doesn't change then it won't have to flush as often.

jonbonazza:
It looks like the bottleneck is, surprisingly, when I reset the verticies at the end of `flush`. If I comment that out, it's super fast (albeit incorrect)

PR #619, opened 2018-12-11
Fix issue with comparing shaders.

jonbonazza:
In renderEntityList.Less, the shader comparison was incorrect. Instead
of using the valuie of the render component's shader (a pointer) it
was instead comparing the address of the pointer variable of the render
component's shader (address of the pointer). This was resulting in the
comparison always being false, even if it was the same shader.


Noofbiz:
Thanks! Great job!

PR #621, opened 2018-12-12
Update demo readme

jtuttle:
Removed the dead Entity link from the demo README and added a section for the Entity Scroller demo.


Noofbiz:
Thanks for the catch!

PR #622, opened 2018-12-12
skip EntityScroller update loop if target entity is nil

### LINKED ISSUE #610: EntityScroller nil pointer safety
If EntityScroller is added without a SpaceComponent set, a nil pointer exception will occur on first update.

Possible fix in Update method:
```go
if c.SpaceComponent == nil{
  return
}
```

jtuttle:
Closes #610


Noofbiz:
Looks great! Thanks for your contribution!

PR #626, opened 2018-12-14
feature: action-based input support

jonbonazza:
This adds basic support for action-based input. In order to keep things
backwards compatible, I needed to make some compromises on naming and
such, but it's not too bad.


jonbonazza:
### Do not merge yet!

I still need to do proper testing of this and also create more unit tests. Some demos might also be useful.

PR #628, opened 2018-12-22
changes to engo so vulkan works properly

Noofbiz:


engo_sdl.go
```diff
@@ -32,7 +34,8 @@ var (
 	Gl           *gl.Context
 	sdlGLContext sdl.GLContext
 
-	scale = float32(1)
+	scale       = float32(1)
+	shouldClose bool
```
jonbonazza: Since we are now using a channel for close notification in our other renderers, we should follow the same pattern here.

Noofbiz: Yeah, but that leads to deadlock since the check to exit is in the update loop due to sdl's poll events. I think there's a callback I can use instead though. I'll look.

Noofbiz: Fixed both SDL and GLFW to use the channel exclusively ^_^!

engo_sdl.go
```diff
@@ -157,8 +130,9 @@ func RunIteration() {
 		for event := sdl.PollEvent(); event != nil; event = sdl.PollEvent() {
 			switch e := event.(type) {
 			case *sdl.QuitEvent:
-				Exit()
-				return
+				go func() {
+					closeGame <- struct{}{}
```
jonbonazza: Why didn't we just use `close(closeGame)`?

Noofbiz: Exit calls close on it already, if it hits that and it's already closed then you'll get a panic: close of closed channel

jonbonazza: Writing to a closed channel will also panic though. Unless I am missing something.

Noofbiz: True. I'll try using sync.Once so it only closes one time in case it's hit multiple times, then I can call exit here safely.

jonbonazza: I think `sync.Once` is a great plan. :)

PR #631, opened 2018-12-31
expose generateFontAtlas in Font

ZwodahS:
This will be useful if you need to generate Font Atlas. Exposing it is no harm

### DIFF
--- a/common/font.go
+++ b/common/font.go
@@ -248,6 +248,12 @@ func (f *Font) generateFontAtlas(c int) FontAtlas {
 	return atlas
 }
 
+// GenerateFontAtlas generates the font atlas for this given font, using the first `c` Unicode characters.
+// This should only be used if you are writing your own custom text shader.
+func (f *Font) GenerateFontAtlas(c int) FontAtlas {
+	return f.generateFontAtlas(c)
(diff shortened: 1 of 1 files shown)


common/font.go
```diff
@@ -248,6 +248,11 @@ func (f *Font) generateFontAtlas(c int) FontAtlas {
 	return atlas
 }
 
+// Should only use this if you are using your own custom text shader.
```
Noofbiz: The comment should start with the function name and describe what it does. Something like "GenerateFontAtlas creates an atlas using the Font of the first c UTF-8 characters. This function should only be used in custom Render systems or Shaders. To create the font atlas for the normal RenderSystem, use Create or CreatePreloaded."

common/font.go
```diff
@@ -248,6 +248,11 @@ func (f *Font) generateFontAtlas(c int) FontAtlas {
...
+func (f *Font) GenerateFontAtlas(c int) FontAtlas {
```
Noofbiz: For brevity I would just capitalize the original function rather than create a separate exported function. That said, the go standard library actually does this often so this should be fine.

PR #636, opened 2019-01-15
view-culling for basic shader

boombuler:
Hi,

after thinking about frustum culling for 2D, I've came up with this simple solution. I'm not 100% sure this is really faster so this needs to be tested.

If this could be optimized, it could be a performance gain. 
the cpu-profiler tells me that it spend 68% of all the `RenderSystem` time in `ShouldDraw` and 86% of that time is spend in `engo.MultiplyMatrixVector`

### DIFF
--- a/common/render.go
+++ b/common/render.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"image/color"
 	"sort"
+	"unsafe"
 
 	"sync"
 
@@ -289,6 +290,11 @@ func (rs *RenderSystem) Remove(basic ecs.BasicEntity) {
 	delete(rs.ids, basic.ID())
 }
 
+// Unsafe interface for instance comparison
+type iface struct {
+	Type, Data unsafe.Pointer
+}
+
 // Update draws the entities in the RenderSystem to the OpenGL Surface.
 func (rs *RenderSystem) Update(dt float32) {
 	if engo.Headless() {
@@ -309,7 +315,7 @@ func (rs *RenderSystem) Update(dt float32) {
 
 	preparedCullingShaders := make(map[CullingShader]struct{})
 	var cullingShader CullingShader // current culling shader
-	var prevShader Shader           // shader of the previous entity
+	var prevShader iface            // shader of the previous entity
 
 	// TODO: it's linear for now, but that might very well be a bad idea
 	for _, e := range rs.entities {
@@ -320,9 +326,12 @@ func (rs *RenderSystem) Update(dt float32) {
 		// Retrieve a shader, may be the default one -- then use it if we aren't already using it
 		shader := e.RenderComponent.shader
 
-		if shader != prevShader {
+		// comparing the instances using unsafe pointers seems to be a little faster (about 0.007 ns on a "slow" machine)
+		// so using this unsafe method to compare shader != prevShader gives a nice performance boost when using many invisible entities.
+		shaderIface := *(*iface)(unsafe.Pointer(&shader))
+		if shaderIface.Data != prevShader.Data {
 			// to increase performance avoid the type assertions when possible
-			prevShader = shader
+			prevShader = shaderIface
 			if cs, ok := shader.(CullingShader); ok {
 				cullingShader = cs
 				if _, isPrepared := preparedCullingShaders[cullingShader]; !isPrepared {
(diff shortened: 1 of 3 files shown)


jonbonazza:
Looks good to me, though I'd like to see some benchmark comparisons before merging, just to make sure it doesn't slow things down

boombuler:
It's not very trivial to create a benchmark for the rendersystem but here are some numbers from the profiler, rendering a tiled map with 430k tile entities.

With this patch:
```
%cum   Name
51.71% engo.io/engo/common.(*RenderSystem).Update
33.58% engo.io/engo/common.(*basicShader).ShouldDraw
 0.21% engo.io/engo/common.(*basicShader).Draw
```

Without this paths:
```
%cum   Name
59.97% engo.io/engo/common.(*RenderSystem).Update
55.47% engo.io/engo/common.(*basicShader).Draw
```

as I said, not very much. But I'm sure there is room for improvement since most time is spend in `MultiplyMatrixVector` which spends most of its time in `runtime.newobject`. I'll have a look if I could get rid of the allocation there.

boombuler:
Update: Now it doubles up my framerate ;)

Noofbiz:
This is due to [this](https://github.com/hajimehoshi/oto/issues/53). You can either just wait until it gets fixed (should be quick!) or you can update the appveyor.yaml to clone oto, `git checkout v0.2.2`, and do a `git reset --hard` to oto. This would be done before engo is cloned. I actually wouldn't do that, though, oto should be fixed soon.

hajimehoshi:
Sorry for the trouble in oto.

Noofbiz:
Sorry about the delay on this, I'm finally getting around to it ^_^! Only thing I noticed is that when the height / width isn't stored correctly in the space component then if the top left corner leaves the screen, the whole tile disappears. This creates blank white spaces on the top and left hand sides of, for example, the adventure demo. It could be fixed by using the render component in ShouldDraw rather than the space component's AABB as follows:

```go
func (s *basicShader) ShouldDraw(rc *RenderComponent, sc *SpaceComponent) bool {
	aabb := engo.AABB{
		Min: sc.Position,
		Max: engo.Point{
			X: sc.Position.X + (rc.Drawable.Width() * rc.Scale.X),
			Y: sc.Position.Y + (rc.Drawable.Height() * rc.Scale.Y),
		},
	}

	min := aabb.Min.MultiplyMatrixVector(s.cullingMatrix)
	max := aabb.Max.MultiplyMatrixVector(s.cullingMatrix)

	return !((min.X < -1 && max.X < -1) || // Both points left of the "viewport"
		(min.X > 1 && max.X > 1) || // Both points right of the "viewport"
		(min.Y < -1 && max.Y < -1) || // Both points above of the "viewport"
		(min.Y > 1 && max.Y > 1)) // Both points below of the "viewport"
}
```

However, I would also like to account for rotation as well in this calculation.

Thanks for all the work! This looks great!

boombuler:
I have not tested rotation but I've looked at the code and it seems like aabb is taking the rotation into account thats why I used it. Once using the drawable height / width, the rotation could be done by matrix transformation. but that could cause additional allocations which might have an impact on the performance. anyone has a good idea on this one?

Noofbiz:
I think making a temporary space component equal to the width / height of the drawable would be best. It gives access to the AABB() and Corners() methods that also account for rotation, so no need to rewrite those functions.

```go
func (s *basicShader) ShouldDraw(rc *RenderComponent, sc *SpaceComponent) bool {
	tsc := SpaceComponent{
		Position: sc.Position,
		Width:    rc.Drawable.Width() * rc.Scale.X,
		Height:   rc.Drawable.Height() * rc.Scale.Y,
	}

	aabb := tsc.AABB()

	min := aabb.Min.MultiplyMatrixVector(s.cullingMatrix)
	max := aabb.Max.MultiplyMatrixVector(s.cullingMatrix)

	return !((min.X < -1 && max.X < -1) || // Both points left of the "viewport"
		(min.X > 1 && max.X > 1) || // Both points right of the "viewport"
		(min.Y < -1 && max.Y < -1) || // Both points above of the "viewport"
		(min.Y > 1 && max.Y > 1)) // Both points below of the "viewport"
}
```

Any other ideas? ^_^

boombuler:
Hi again,

I've implemented your proposed changes. and now also taking all four corners for checking, since it made some problems with camera rotation.

Noofbiz:
Thanks for all the hard work! Looks awesome!

PR #640, opened 2019-02-05
fixed shader so hud is unaffected by camera

### LINKED ISSUE #632: Hud shader not working on the traffic manager demo with latest version of engo
For some reason the background of the hud for the traffic manager tutorial isn't working correctly. It's just sitting in the spot it starts in and moves as if it's using the camera even though it's set to use the hud shader. It works correctly on v1.0 but not on master.

Noofbiz:
fixes #632


Noofbiz:
TravisCI switched to using GO111MODULE=on, so I now have to fix all that is what's happening :ghost:

PR #642, opened 2019-02-06
js testing using wasm

### LINKED ISSUE #641: Gopherjs or wasm testing
Now that travis defaults to $GO111MODULES=on and gopherjs is not module aware, the gopherjs tests have been disabled. This could easily be fixed when gopherjs becomes module aware, but until then it could be fixed by either disabling go modules in the tests or testing instead using wasm and a framework like [puppeteer](https://github.com/GoogleChrome/puppeteer).

Noofbiz:


Noofbiz:
annnnnnd they changed it back to GO111MODULES=auto >_>

PR #643, opened 2019-02-07
Added a `Shader` func to access the shader

boombuler:


Noofbiz:

PR #644, opened 2019-02-13
prevent allocations on MultiplyMatrixVector

boombuler:
due to a lot of calls in `ShouldDraw` all those allocations of the float slices trigger the GC more often then needed. 

this change reduces the allocation from 3 slices per call to 0.


Noofbiz:
Looks great, thanks!

Issue #679, opened 2019-08-17
Sprite "random-lines" when zoom-out/move


Inkeliz:
I tested the "TrafficManager" demo, then I notice one bug when move (using the mouse at the edge of the screen) or zoom-out (using the mouse wheel). However, that bug ONLY happens against "sprites".

I use the following code to reproduce that issue:

	grass := make([]*Grass, 0)

	// Using image
	image, _ := common.LoadedSprite("tiles/mapTile_077.png")
	for x := float32(0); x < 512; x += 64 {
		for y := float32(0); y < 512; y += 64 {
			tile := &Grass{BasicEntity: ecs.NewBasic()}
			tile.SpaceComponent.Position = engo.Point{X: x, Y: y}
			tile.RenderComponent.Drawable = image
			tile.RenderComponent.SetZIndex(1)
			grass = append(grass, tile)
		}
	}

	// Using Sprite
	sprite := Spritesheet.Cell(50)
	for x := float32(0); x < 512; x += 64 {
		for y := float32(512); y < 1024; y += 64 {
			tile := &Grass{BasicEntity: ecs.NewBasic()}
			tile.SpaceComponent.Position = engo.Point{X: x, Y: y}
			tile.RenderComponent.Drawable = sprite
			tile.RenderComponent.SetZIndex(1)
			grass = append(grass, tile)
		}
	}

That is the result:

![image](https://user-images.githubusercontent.com/4644332/63215162-83705b80-c11a-11e9-823c-00aaa32c8056.png)

When I use the image, there are no such "lines" between these images. Using "images": I'm able to move, zoom and zoom-out without issue. However, the "sprites" display some "lines" at random.

-------------------

The full code:

```
package main

import (
	"bytes"
	"github.com/EngoEngine/ecs"
	"github.com/EngoEngine/engo"
	"github.com/EngoEngine/engo/common"
	"golang.org/x/image/font/gofont/gosmallcaps"
	"image/color"
	"path/filepath"
	"strings"
)

const (
	KeyboardScrollSpeed = 400
	EdgeScrollSpeed     = KeyboardScrollSpeed
	EdgeWidth           = 20
	ZoomSpeed           = -0.125
)

var Spritesheet *common.Spritesheet

type Grass struct {
	ecs.BasicEntity
	common.RenderComponent
	common.SpaceComponent
}

type myScene struct{}

// Type uniquely defines your game type
func (*myScene) Type() string { return "myGame" }

[...]

PR #680, opened 2019-08-18
Fix filters

Inkeliz:
That change is intended to fix https://github.com/EngoEngine/engo/issues/679. The default value of `lastMagFilter` and `lastMinFilter` are equal to `FilterNearest`, which seems never to apply that filter.

That changes set a default value to 255 (_uint8 maximum value_), which will be not equal to 0 by default:
```
s.lastMagFilter = 255
s.lastMinFilter = 255
```

However, any ZoomFilter still `FilterNearest` by default (since it still `0`), any existing code still work without any change.

---------

We can probably add the `255` as a new `ZoomFilter` at `render.go`:
```
const (
   // FilterNearest is a simple nearest neighbor algorithm
   FilterNearest ZoomFilter = iota
   // FilterLinear is a bilinear interpolation algorithm
   FilterLinear
   // filterNone is a default value for render functions
   filterNone ZoomFilter = 255
)
```

I try to make changes as minimum as possible, so I choose not to create a new constant. Maybe have a better fix for that issue.


Noofbiz:
The change is great. I missed this bug when testing the filters, but looking back it's exactly what caused a few issues with them lol. Thanks for the fix! I'll work on the CI stuff and get it in ^_^

Inkeliz:
I'm still testing some stuff and trying to understand it. Seems that `lastMagFilter` and `lastMinFilter` is never updated. I imagine that something like the code below was suppose to exist:

````
if s.lastMagFilter != ren.magFilter {
   //...
   s.lastMagFilter = ren.magFilter
}
````

The current code will be always `255`. However, if that is suppose to update, we need to reset the value when change the texture:

	if s.lastTexture != ren.Drawable.Texture() {
		s.flush()
		engo.Gl.BindTexture(engo.Gl.TEXTURE_2D, ren.Drawable.Texture())
		s.lastTexture = ren.Drawable.Texture()
		s.lastMinFilter = 255 // << Here
		s.lastMagFilter = 255  // << Here

I think create a new `func (s *basicShader) reset()`, to set a default value, maybe is better.

-----------

I notice that my change cause a significant performance impact, from 2% to 7% of GPU usage. I remove the `s.flush()` inside each `if s.lastRepeating != ren.Repeat`, `if s.lastMagFilter != ren.magFilter` and `s.lastMinFilter != ren.minFilter`. 

I don't understand anything about OpenGL, or graphical stuff, so that can be wrong. But, I don't see the point of the flush here, seems that the filter will change on the fly, during the render, I don't know, I need to sleep now. :P

Inkeliz:
The [previous change](https://github.com/EngoEngine/engo/pull/680/commits/e351fc6f4597c7eb6dec801c8cb4aba038cf2e1a) causes the Engo to call `engo.Gl.TexParameteri` everytime, since the `lastRepeating`, `lastMagFilter` and `lastMinFilter` was never updated. The `if s.lastMagFilter != ren.magFilter` is always `true`.

That new change adds a new `s.lastMagFilter = ren.magFilter` and so on, reducing calls to OpenGL. I keep the `s.flush()`, but it only applies if the filter is different from the previous one, _as expected_. 

I create a new function `setTexture` to set all `s.last*` stuff to 255.

Noofbiz:
Sorry it's taking so long to get this properly reviewed and put in. I've been having a combination of personal and medical issues this year which has really limited my free time. It looks good looking it over so I'll try to get it in hopefully by tomorrow.

Issue #681, opened 2019-08-19
Camera moves without interaction at Fullscreen 


Inkeliz:
There's a bug at fullscreen mode, that didn't set Width/Height correctly. I could reproduce that issue using the TrafficManage with one change:

Replacing:

	opts := engo.RunOptions{
		Title:          "TrafficManager",
		Width:          800,
		Height:         800,
		StandardInputs: true,
	}

By:

	opts := engo.RunOptions{
		Title:          "TrafficManager",
		Fullscreen:     true,
		StandardInputs: true,
	}

That small change causes the camera to move without move the cursor to the edge, the camera will keep move until reach the end.

It's possible to fix using the `glfw` directly, but I think Engo must do that itself:

    glfw.Init()

	monitor := glfw.GetPrimaryMonitor()
	vid := monitor.GetVideoMode()

	opts := engo.RunOptions{
		Title:          "Test",
		Width:          vid.Width,
		Height:         vid.Height,
		Fullscreen:     true,
		StandardInputs: true,
	}

I'm looking into the Engo code to find the bug.

Inkeliz:
Seems that the problem is related to `gameWidth` and `gameHeight`, at that line:

https://github.com/EngoEngine/engo/blob/6f88a7379a25d81e976cf6e86ea3617f3b480b1a/engo_glfw.go#L75-L84

_The order seems wrong._

-----------

That fixes the issue:

	if fullscreen {
		width = mode.Width
		height = mode.Height
		glfw.WindowHint(glfw.Decorated, 0)
	} else {
		monitor = nil
	}

	gameWidth = float32(width)
	gameHeight = float32(height)

Should I submit a PR?

Noofbiz:
Sounds great to me. Pull requests are always welcome!

PR #684, opened 2019-11-14
fix width/height at fullscreen

Inkeliz:
That change fixes the issue #681.


Noofbiz:
Awesome thanks for your contribution!

PR #688, opened 2019-12-13
Change z-index to public

Inkeliz:
I'm not sure why, but the current version of Engo has a `zIndex` as private value. 

The only way to set ZIndex is by using `SetZIndex`, at the current version. That makes it impossible to set the value in advanced (at the `&RenderComponent{}`). In my case doesn't make any sense to set the ZIndex using the `SetZIndex`, since some elements have always the same ZIndex (like: background is 0 and player is 1).

I'm not aware of any issue of that change, but I don't know what wasn't public before.

### DIFF
--- a/common/render.go
+++ b/common/render.go
@@ -1,15 +1,13 @@
 package common
 
 import (
-	"fmt"
-	"image/color"
-	"sort"
-
-	"sync"
-
 	"github.com/EngoEngine/ecs"
 	"github.com/EngoEngine/engo"
 	"github.com/EngoEngine/gl"
+	"image/color"
+	"sort"
+	"sync"
+	"unsafe"
 )
 
 const (
@@ -80,13 +78,15 @@ type RenderComponent struct {
 	// BufferContent contains the buffer data
 	// Avoid using it unless your are writing a custom shader
 	BufferContent []float32
-	// ZIndex defines the order which the content is drawn to the screen. Higher z-indices are drawn on top of
-	// lower ones if men overlap.
-	ZIndex float32
+	// StartZIndex defines the initial Z-Index. Z-Index defines the order which the content is drawn to the
+	// screen. Higher z-indices are drawn on top of lower ones. Beware that you must use `SetZIndex` function to change
+	// the Z-Index.
+	StartZIndex float32
 
 	magFilter, minFilter ZoomFilter
 
 	shader Shader
+	zIndex float32
 }
 
 // SetShader sets the shader used by the RenderComponent.
@@ -125,7 +125,7 @@ func (r *RenderComponent) Shader() Shader {
 // SetZIndex sets the order that the RenderComponent is drawn to the screen. Higher z-indices are drawn on top of
 // lower ones if they overlap.
 func (r *RenderComponent) SetZIndex(index float32) {
-	r.ZIndex = index
+	r.zIndex = index
 	engo.Mailbox.Dispatch(&renderChangeMessage{})
 }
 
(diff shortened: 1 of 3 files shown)


Noofbiz:
It does this because it has to set rs.sortingNeeded so that the system can sort the textures properly. If it's public and you set it yourself, it may not actually do anything, since the textures still draw in the same order every pass.
I do see what you mean, though. I think maybe there can be another public variable, such as `StartZIndex`, that sets the starting Z-Index. Then near the top of the render system's Add function, you just call `SetZIndex(render.StartZIndex)`. Would that solve your problem?

Inkeliz:
Yes, I don't thought that someone will set `something.ZIndex = 3` after already use the `render.Add(...)`. 😅 

The `StartZIndex` is better.

common/render.go
```diff
@@ -153,63 +155,46 @@ func (r renderEntityList) Len() int {
...
-				}
-				return t1 < t2
-			}
-		}
+	if r[i].RenderComponent.zIndex != r[j].RenderComponent.zIndex {
+		return r[i].RenderComponent.zIndex < r[j].RenderComponent.zIndex
+	}
+
+	// TODO: optimize this for performance (is it enough? maybe "cache" shader uintptr (i.e: r.shaderPtr)?)
+	p1, p2 := (*[2]uintptr)(unsafe.Pointer(&r[i].RenderComponent.shader))[1], (*[2]uintptr)(unsafe.Pointer(&r[j].RenderComponent.shader))[1]
```
Noofbiz: Actually, there's already a shader comparison function `compareShaders` that does that during the draw step. It's optimized for speed and also has a version that works for gopherjs and wasm. We should just use that here XP

Inkeliz: The _problem_ is that `compareShaders` just return a bool, but not the `pointer` (we need to compare `p1 < p2`). Maybe a new `compareShadersLess(a, b Shader) bool` 🤣. In that case if it's true it's not equal and just return.

```
p := comapreSharersLess(r[i].RenderComponent.shader, r[j].RenderComponent.shader)
if p {
    return p
}
```
```
func compareShadersLess(a, b Shader) bool {
	aIface := *(*iface)(unsafe.Pointer(&a))
	bIface := *(*iface)(unsafe.Pointer(&b))
	return uintptr(aIface.Data) < uintptr(bIface.Data)
}
```
😅

Noofbiz: Alright, but the way it's done now breaks gopherjs and wasm builds. To fix this I propose we add to render_notjs.go a `lessShaders(a, b Shader) bool` that will return wether shader a is less than shader b using the unsafe package, and then a similar function in render_js.go that does the same thing without using unsafe. Then call that at the return for `p1 < p2`.

We can do this in one of two ways:
1) I can merge this as-is and then fix it in a later commit myself

2) You can fix it here, if you have time and want to do it

either way works for me ^_^

Inkeliz: I think I fix that. But has another issue. I'm testing the `entityscroller.go`, it doesn't render because of `StartZIndex`. I don't know why, but `render.zIndex = render.StartZIndex` breaks the game. The only way to fix seems to be:

```
func (r *RenderComponent) SetZIndex(index float32) {
	r.StartZIndex = index
	r.zIndex = index
	engo.Mailbox.Dispatch(&renderChangeMessage{})
}
```

However, how the same component is added again? 😕

Inkeliz: Ok. I found the issue.

Inkeliz:
I think it's fine now, the rendering is working as expected. I didn't test on GopherJS. I'm going to sleep now. 😅

Noofbiz:
Looks good, thank you for your contribution!

PR #692, opened 2020-01-07
fix memory leak in Demo "falling"

aligator:
This PR fixes a memory leak. Each stone adds itself as references to the RenderSystem, FallingSystem and CollisionSystem. All references of them were never removed.
The remove-functions were implemented (not completely...) but never called.

So I extended it to remove all corresponding references and call it when the stones reach the bottom.

### DIFF
--- a/demos/falling/falling.go
+++ b/demos/falling/falling.go
@@ -1,5 +1,3 @@
-//+build demo
-
 package main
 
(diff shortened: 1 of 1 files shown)


Noofbiz:
Awesome! Thanks for your contribution!

Issue #694, opened 2020-01-11
Camera LongTasks seems broken


Inkeliz:
Since my [workaround to fix texture](https://github.com/EngoEngine/engo/issues/691#issuecomment-572315787), I create some `Steps []float64`, so I can use:

```
m.MouseZoomer.Steps = []float32{1, 2, 4, 8}
```

However, the change of zoom and angle is very roughly. In way to prevent it, I make one `m.MouseZoomer.Duration = time.Second`. The `Duration` supplies the `Duration` of `CameraMessage`. 

It was suppose to take 1 second to change the zoom from 1 to 2, another 1 second to change 2 to 4.

-------------

However, there's some bugs:

https://github.com/EngoEngine/engo/blob/db964088800cfff0d9f08710dd4391607189a53c/common/camera.go#L171

That doesn't work. The `time.Duration` is `int64`. If the game is running at 60 FPS the `dt` will be 0.16 seconds, which will be 0, since `int64(float32(0.16))` is 0. It creates a infinite loop. I think the solution must be:

		const second = float32(time.Second)
		longTask.Duration -= time.Duration(dt * second)

Then, have another problem, which seems to be caused by `longTask.speed = longTask.Value / float32(longTask.Duration.Seconds())` when `longTask.Incremental = false`. It should be `false` since it is a "steped" zoom.

---------------

I'm trying to fix that, but I think it will be very different from the current implementation. 😐

Inkeliz:
I found the second issue. It's kinda related to `longTask.speed = longTask.Value / float32(longTask.Duration.Seconds())`. The problem is that the value isn't exactly, for instance you can set to zoom-out 2, but it only zoom-out 1.99. 

The fix for that is enforce to set the correct value when `longTask.Duration <= time.Duration(0)`. So, the last step will `zoomTo(2)`.

I'll try to make an pull-request, because I blow-up my fork. 😅

PR #695, opened 2020-01-12
fix camera longtask

Inkeliz:
It fixes https://github.com/EngoEngine/engo/issues/694. It's untested, I just port from my fork. I just create a new function to warp the `switch` just to avoid a new  `switch`. 🤣


Noofbiz:
Awesome! Thanks for the help!

Issue #538, opened 2018-04-30
JustReleased misbehavior on Ubuntu family 


nvma:
Hey, guys. I've been surprised to find out that it has been a two years (wew) since I first encountered this issue, back then we've discussed it on gitter and unfortunately nothing came out of it, I've poked around for some time but was unable to resolve it on my own.
Since then I've change several workstations and Ubuntu 18 has been released, I thought I might try again, but unfortunately the issue is still there and it's really upsetting.

**To summarize the problem**: whenever keyboard button goes through Press/Release state, right after Release it repeats the Press/Release cycle. 
I've recorded a short video to demonstrate the problem, running latest adventure demo: https://webm.xyz/webm/5ae7009561f4bc000f1671f8
With this behavior it's impossible to make any sort of keyboard controls. 
Also the following scenario doesn't work at all:
LeftButton is pressed -> we move left.
RightButton is pressed while LeftButton is pressed -> we move right
RightButton is released while LeftButton is pressed -> we should back to moving left, but we stop.

When I first encountered this issue I had an Arch machine around and I was not able to reproduce this issue there. The issue persisted through 16, 17 and now 18 Ubuntu versions.

I wasn't able to find anything in the issue tracker related to this problem and it makes me wonder whether I'm the sole Ubuntu user, which seems to be unlikely. But anyways I would really appreciate any tips that might shed light on this problem.

Thanks a lot!

Noofbiz:
Hello @nvma! I've got a fix that works for me here https://github.com/EngoEngine/engo/tree/just-released-bug if you want to check it out

nvma:
@Noofbiz hey, sorry that couldn't react earlier. Unfortunately no impact on my end running the latest adventure demo. 
Would you please tell me what happens on your machine when you press, say, 'D' for a ten or so seconds, when the character has already reached the edge of the world, and then release it? Does animation immediately stop?
Because for me, the longer you keep the button in pressed state, the longer it will be 'pressed' again after I release it.

Noofbiz:
It appears to be a bug with glfw itself and is present in the C version as well [here] (https://github.com/glfw/glfw/issues/747). Looks like the fix is actually in, but is in glfw 3.3, which is currently in beta. I've got an issue here #531 to upgrade to it when it's ready. That should fix this issue. Thanks for your help and reporting!

nvma:
[...]

PR #700, opened 2020-02-21
update to glfw 3.3.2

### LINKED ISSUE #531: Upgrade to glfw 3.3 when it's out of beta
glfw 3.3 is at around 90% currently, and provides some nifty new features, such as support for vulkan (as well as molten for iOS and OSX) as well as Wayland, which Ubuntu is switching to from X.

Noofbiz:
fixes #531


Noofbiz:
also fixes #538 ^_^

PR #702, opened 2020-03-02
Keep mouse drag when mouse action is neutral

m0re4u:
Previously, when the mouse action was neutral, the `Dragged` state of the mouse was reset to `false`. This change fixes it by keeping it set to `true` as long as one of the mouse buttons is down, requiring a mouse release before the `Dragged` state is exited.


Noofbiz:
Awesome! Thanks for your help!

PR #703, opened 2020-03-05
Fixed missed case in setting Dragged flag

m0re4u:
Missed a case when fixing the `Dragged` flag for the mouse. In this case, the mouse was Clicked but unmoved, which would register both as a click and a drag. With this fix, if the mouse is pressed down and stays in the same place, it is considered to be `Clicked`. If it moves at some point while still pressed, the mouse will be `Dragged ` as long as there is no mouse release.


Noofbiz:
Awesome thanks for the catch. Could probably use this to clean up the mouse demos down the road.

PR #709, opened 2020-04-23
fix create object TextureResource with filled field url

ilyar:


Noofbiz:
Nice catch! Thanks for the contribution!

Noofbiz:
Apologies for the Travis stuff it's always a curmudgeon.

PR #717, opened 2020-05-11
Add support for Texture Atlases (spritesheets)

### LINKED ISSUE #707: Utilize a Sprite Atlas
Now that batch rendering is complete (thanks to everyone for all the help there!) the next step for picking up speed is utilizing a sprite atlas. This reduces texture switching, which reduces draw calls even further! My two cents is we should have a function `common.LoadToFontAtlas(urls ...string) error` that appends the files in the url to a texture which would be our font atlas. It would basically be a `common.SpriteSheet` except rather than use a map[int]Texture it would have a map[string]Texture where the string would be the url. It would re-use the atlas, so if another call is made to `common.LoadToFontAtlas(url)` it would append the new sprites to the current atlas. It would also be a slice rather than single, that way if the textures get bigger than 2048x2048 (or whatever multiple of 2 OpenGL will allow :P) we can start a second one.

otraore:
- Add TextureAtlas loader to the File Loader based on xml exports from [TexturePacker](https://www.codeandweb.com/texturepacker)
- Make it the default loader for xml files
- Use a singular `ImageLoader` object for `engo.Files` to save memory allocations
- Add Viewport to `TextureResource` and use it if defined in `LoadedSprite(..)`
     - Allows the use of the singular sprite sheet texture because singular sprites will be retrieved from the viewport of it

I tried to make it so that users can continue using `common.LoadedSprite` regardless of if they're using a `TextureAtlas` or not. This means that the texture is added to the image loader manually and is referenced using the file url for it.

For example: 
`<SubTexture name="enemyGreen1.png" x="425" y="552" width="93" height="84"/>`

That sprite can be retrieved using 

`texture, err := common.LoadedSprite("enemyGreen1.png")` even though it wasn't loaded from that location in the file-system but loaded from the singular sprite sheet image defined in the xml file.

Tested using: [these files](https://github.com/otraore/space-shooter/tree/master/assets/spritesheets)
I am open to suggestions

In progress: needs review, tests
Closes #707

### DIFF
--- a/common/render_filetype.go
+++ b/common/render_filetype.go
@@ -22,6 +22,7 @@ import (
 	"github.com/EngoEngine/gl"
 )
 
+// imgLoader is the shared imageLoader for all image file formats
 var imgLoader *imageLoader
 
 // TextureResource is the resource used by the RenderSystem. It uses .jpg, .gif, and .png images
--- a/common/render_filetype.go
+++ b/common/render_filetype.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"image"
 	"image/draw"
+
 	// imported to decode jpegs and upload them to the GPU.
 	_ "image/jpeg"
 	// imported to decode .pngs and upload them to the GPU.
@@ -21,12 +22,15 @@ import (
 	"github.com/EngoEngine/gl"
 )
 
+var imgLoader *imageLoader
+
 // TextureResource is the resource used by the RenderSystem. It uses .jpg, .gif, and .png images
 type TextureResource struct {
-	Texture *gl.Texture
-	Width   float32
-	Height  float32
-	url     string
+	Texture  *gl.Texture
+	Width    float32
+	Height   float32
+	Viewport *engo.AABB
+	url      string
 }
 
 // URL is the file path of the TextureResource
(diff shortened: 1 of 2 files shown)


Noofbiz:
Looks great to me. I think integration with texture packer is awesome. Thanks for the awesome features! Down the road I'm thinking we could add an internal Atlas that spends individual sprites to it, so end users can just have a bunch of pngs and not have to worry about texture packer themselves unless they want to, but that's a lot of work and planning that can be done down the road, if it's even necessary at all lol

otraore:
You mean like caching all loaded images into a singular `Texture` internally for performance improvements? If so I don't think that'd work well because the point of the packers are to reduce the size as well. Loading full pngs/jpegs etc will still be at full size. If we were to do the job of packer we'd also try to reduce the size of the images and compress them but at that point there will be image quality loss. I think it's better if we present them both options and let them choose whatever works for them. Let me know what you think.

I made it a goal to keep the same API for loading in singular sprites and entire sprite atlases so that if they decide to use one or the other down the line they only have to change the Load function. 

Thank you for the review though, I think I want to add tests for the loader either in this PR or in a separate one (because writing tests is not fun). Hopefully I'll get to it this week and just do it in this one.

otraore:
Another option for what you want to do would be to use something like [this library](https://github.com/RaniSputnik/lovepac) and have a flag like `PackAssets: true` in `engo.RunOptions` which will pack assets for them.

otraore:
I removed it, I'll see if I can add tests and maybe a demo in another PR. 

Thank's for the review.

Issue #719, opened 2020-06-02
undefined: oto.NewPlayer


waltrapa:
$ go version
go version go1.14.3 linux/amd64

At try to build https://github.com/EngoEngine/engo/tree/master/demos/hello-world fail:
$ go build hello.go 
$# github.com/EngoEngine/engo/common
../../../src/github.com/EngoEngine/engo/common/audio.go:93:21: undefined: oto.NewPlayer

oto example https://github.com/hajimehoshi/oto/blob/master/example/main.go build and run perfect.

Asday:
https://github.com/hajimehoshi/oto/issues/76

otraore:
Fixed via #720. Thanks for bringing attention to the issue! Let us know if you're still having any issues.

PR #720, opened 2020-06-02
Fix audio system by using updated Oto API

otraore:
- Remove deprecated `oto.NewPlayer()` for the `oto.NewContext()` => `context.NewPlayer()`api
- Update doc for `NotResizeable` to be more accurate since it's false by default
- Update `go.mod` for June 2, 2020


otraore:
Fixes #719

PR #723, opened 2020-06-11
Vulkan rendering [WIP] [DO NOT MERGE]

Noofbiz:
Adds the VulkanRenderSystem to `common` behind the build tag `-vulkan` so if people want to build their games with Vulkan instead of OpenGL as the renderer they can ^_^!

This is a work in progress, and is not ready yet. I'm just publishing it as I go like this to get feedback if anyone wants to, and maybe it'll help keep me more committed to doing it xP

TO DO:
- [X] engo updated to fully accommodate Vulkan
- [x] render system split into OpenGL and Vulkan behind build tags 
- [x] file loader working with Vulkan
- [ ] shaders working in Vulkan
- [ ] demos working in Vulkan
- [ ] Vulkan added to the tests in GitHub actions


Noofbiz:
Now at this point I'm thinking some point down the road merging engo_glfw with engo_vulkan. engo could always create the device and swap chain, then it's up to the renderer to use it or not. This could be useful such that we could always default to Vulkan and use OpenGL as a fall-back if the device doesn't support Vulkan. The downside to this, however, is that it would require developers to have Vulkan installed to build engo. Just something to think on, really.

Noofbiz:
Got caught up trying to fix our webgl implementation and had an idea about putting all the vulkan stuff in 'EngoEngine/gl' rather than here. It really is a replacement for gl, we can just mimic opengl functions in vulkan.

PR #725, opened 2020-06-18
no longer js.Value compare using ==

Noofbiz:
syscall/js no longer supports using straight == to compre js values.


Noofbiz:
I'm going put out a new release before upgrading the tests to go1.14.x ^_^

PR #730, opened 2020-06-21
remove separate files for batch size defaults

otraore:
separate files for one function is a bit overkill

Just an enhancement to #728 because `common` has a lot of files so creating new ones should be more meaningful. Let me know if there are other alternatives but this should work


Noofbiz:
Great point, especially since common is already quite intimidating.

PR #731, opened 2020-06-28
Input key modifiers

otraore:
Something to note is that SDL supports modifiers being identifiable by their position on the keyboard. So left shift vs right shift. GLFW as far as I know, does not. To keep the API simple and not requiring the end users to use bitwise operations to check their Modifiers, I just did the checks within the game loop so the API is the same for both platforms. Going to be adding js support in an upcoming commit as well.

ex:
```go
if engo.Input.Modifier == engo.Shift {
    log.Println("is shift")
}
```

### DIFF
--- a/engo_glfw.go
+++ b/engo_glfw.go
@@ -175,6 +175,8 @@ func CreateWindow(title string, width, height int, fullscreen bool, msaa int) {
 		} else if a == glfw.Release {
 			Input.keys.Set(key, false)
 		}
+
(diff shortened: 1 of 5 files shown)


otraore:
Hey @Noofbiz can you "squash and merge" instead of just doing a regular merge? It's just to keep the main commit messages cleaner without all the working commits in the PR. Thanks, appreciate it for getting this merged in.

Issue #738, opened 2020-08-05
Can't build with "headless" build tag 


kori-irrlicht:
I tried to build my project with the headless tag, but got the following error:
github.com/!engo!engine/engo@v1.0.5/engo_empty.go:79:4: fallthrough statement out of place

Noofbiz:
Hmm it looks like a fall through doesn't work with select, maybe the compiler used to build it anyway? I know I've built a headless app to make sure it worked at some point lol

All it needs is the fall through part run on it's own goroutine like it does for the glfw version. Not sure why it doesn't already though

Noofbiz:
I really hoped go built smarter and only the packages actually used. That way just having that headless variable check would take care of it.  

There’s a few options. You can use your headless build of oto, maybe even submit a PR to oto so it can be put in the project. I can see lots of uses for it, such as testing and building on servers. 
Another option is to use a dummy driver on the server. This would be very useful for testing, since you can get the byte output of the sound driver to ensure you’re outputting what you’re inputting. Your case I don’t believe it offers any advantages over the headless version.
The last option is to split common/audio.go into common/audio_headless.go and common/audio.go and use build tags to ensure oto is never touched when you pass the headless tag to engo. I think this one might be good just so that next time someone tries to build with the headless tag it works as expected.

PR #739, opened 2020-08-06
Removed fallthrough case since it caused build errors

kori-irrlicht:
I removed the fallthrough statement. Otherwise it already looked like the glfw version, so I think this should be the necessary changes to fix the compile error I received in #738 

There is still another problem with the headless build: The oto dependency is causing trouble, if no drivers are installed on a build system. But I'm not sure if this should be fixed in this pull request (or even in this repository)

### DIFF
--- a/engo_empty.go
+++ b/engo_empty.go
@@ -75,8 +75,6 @@ func runLoop(defaultScene Scene, headless bool) {
 		case <-resetLoopTicker:
 			ticker.Stop()
 			ticker = time.NewTicker(time.Duration(int(time.Second) / opts.FPSLimit))
-		case <-c:
-			fallthrough
 		case <-closeGame:
 			ticker.Stop()
 			closeEvent()


Noofbiz:
The failures are not due to anything you did, and the PR looks good to me. Also, fixing oto should be relatively easy, just have to use the same method we do for running tests. Try building it with engo.RunOptions has Headless = true. That should skip over oto and use a fake readwriteseeker instead of oto

Noofbiz:
Thanks for the help! Let me know if the headless option works!

kori-irrlicht:
Thanks for the hint, but I was already using the headless option and the error occurs while building the project. Would it be possible to skip the oto dependency this way?
I can build it on my machine, but building it on a build server causes an error, because it can't find the alsa package required by the Linux drivers of oto.

For now, I created a workaround by forking oto and adding an empty driver, which will ignore everything. It works for me, but I'm not sure if this is the best solution overall. 

The error message:
```# pkg-config --cflags  -- alsa
Package alsa was not found in the pkg-config search path.
Perhaps you should add the directory containing 'alsa.pc'
to the PKG_CONFIG_PATH environment variable
No package 'alsa' found
pkg-config: exit status 1
```
If it would help, I can also post a link to my project

PR #757, opened 2021-02-12
Added tile animation to tilmap demo

raziel2244:
Added tilemap animations to TMX file and implemented them in the tilemap demo.


Noofbiz:
Looks great! Thanks for adding this!

raziel2244:
Missed it off my feature PR so happy to add it

PR #759, opened 2021-02-17
Added Go embed demo (Go 1.16) [WIP] [DO NOT MERGE]

eth0net:
Added a demo to show how to embed assets using the new Go embed package.

Does require go.mod to be set to Go 1.16 so this should probably be held back for a while.


raziel2244:
Ah yeah forgot those would need updating, updating those now.

raziel2244:
Okay so still not passing... think I'll just keep this aside for later. Once 1.16 is more established and Engo maybe gets a go.mod update for some reason besides this it can be looked into again.

For anyone looking to embed using Go embed, examples can be provided elsewhere worst case.

raziel2244:
Reopened as WIP to keep it tracked here for the future.

Noofbiz:
I fixed the traffic manager issues, so they should pass if they re-run. The only issue I see is I think engo's go mod version should stay 1.14. Even though 1.16 is required for the embedding, that's not a feature engo itself uses so it shouldn't need to increase the version for it.

go.mod
```diff
@@ -31,4 +31,4 @@ require (
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
 
-go 1.14
+go 1.16
```
Noofbiz, line 34: This shouldn't need to be updated, only testing

Noofbiz (changes requested):

raziel2244:
Thanks, yeah I was thinking it's probably best to leave the go version for Engo as 1.14. Unfortunately, since the demo is within the module it fails to build without the update to go.mod.

raziel2244:
Just looking to see if there is a way to set the -lang flag for build as that could be a way round it, maybe.

raziel2244:
Can't see a way around it for now, unfortunately. The demo can live elsewhere in the meantime though, such as just in the README file.

Noofbiz:
Alright, got the CI stuff working. Turns out the go.mod includes the demos, but it's not a hard/fast rule. I think 1.16 is mature enough at this point to move up to it if you want to merge this in. Looks good to me!

Noofbiz (approved):

eth0net:
Ah well if you're happy to up the Go version then go ahead with the merge, I believe this is all ready!

PR #770, opened 2021-03-18
added modules to all the demos

Noofbiz:


Noofbiz:
Should fix the issues with #759 requiring a higher version of go

PR #780, opened 2021-08-04
Add rotation based on TMX flipping flag

### LINKED ISSUE #779: TMX resources don't respect transformations when loading tiles
Thanks again for this awesome engine! When trying to build a road segment I found that rotated tiles are not rotated in-game.

In Tiled you can rotate tiles (with `z`) and it also stores it correctly, in the file :
```
278,2684354838,3221225750,1610613014,
```
those are 4x the same tile but rotated (0º, 90º, 180º, 270º).

This is how it looks in Tiled:
Screenshot 2021-08-02 at 18 18 35

This is how it looks in-game:
Screenshot 2021-08-02 at 18 19 46

EDIT:
This seems to be the location where the `Flipping` attribute gets lost and should be applied to the image?

https://github.com/EngoEngine/engo/blob/d6f9c6aafb3d9124495edf7115b1b2fb0c35b9c8/common/tmx_level.go#L256-L261

EDIT2:
this is the way Tiled calculates the rotation: https://github.com/mapeditor/tiled/blob/ba252f2bb1666510a2cf3851d105db443968f37c/src/libtiled/maprenderer.cpp#L383-L410

My idea would be to calculate the rotation in the conversion from TMX tile to an Engo tile and store the rotation in degree on the Engo tile.

EDIT 3:
Another option might be to pass on the individual flags `isFlippedDiagonally`, `isFlippedHorizontally`, `isFlippedVertically` and leave it up to the implementer to work with it.

jan-xyz:
When working on my project I noticed that Engo is losing the information about flipped tiles in the tile map when converting from the TMX structure to the internal model. This pull request adds this feature by the means of adding a Rotation in degree to the internal TMX Tile model. I just now went this path, if this is in some way not how you intend to solve this problem, please let me know 🚀  I hope this is also in your interest and thank you for this really fun 2D engine!

This adds also a changed map to the `adventure` demo with 4 rotated flowers on the ground.
Screenshot 2021-08-05 at 01 21 44
fixes #779

### DIFF
--- a/common/level.go
+++ b/common/level.go
@@ -281,4 +281,6 @@ type Tile struct {
 	Image     *Texture
 	Drawables []Drawable
 	Animation *Animation
+	// Rotation of the Tile in degrees
+	Rotation float32
 }
--- a/common/tmx_level.go
+++ b/common/tmx_level.go
@@ -257,6 +257,7 @@ func (l *Level) unpackTiles(x, y, w, h int, d []tmx.Data) []*Tile {
 				X: float32(x),
 				Y: float32(y),
 			}))
+			tile.Rotation = convertFlipToRotation(t.Flipping)
 			ret = append(ret, tile)
 			l.pointMap[mapPoint{X: x, Y: y}] = tile
 			switch l.RenderOrder {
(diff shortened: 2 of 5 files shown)


Noofbiz:
Looks great! Thank you for your contribution!

PR #796, opened 2023-07-19
chore: remove refs to deprecated io/ioutil

testwill:


Noofbiz:
Looks good! Thanks for going through all that! I'll see what's up with the tests and get the ball rolling.

PR #805, opened 2024-06-20
Correctly build gamepads_empty.go for mobile.

brunoga:
- This is required for newer Go versions.
- Help with compiling code for mobile.


brunoga:
@Noofbiz any chance you can take a look at this one? This seems to be the last thing preventing code to compile for both android and iOS.

Noofbiz:
The commas appear to be the issue. I think removing them again will make the builds succeed. Right now it's saying the new/old way don't match. If you get that in and it passes I'll push them through. 

Thanks!

